PR #8, opened 2019-02-12
Configure Bootstrap to use the JAC's brand colours

ollietreend:
This PR replaces the CDN version of Bootstrap with a customised build which uses the JAC's official brand colours.

By enabling support for SCSS assets and including the official `bootstrap` npm package, we can build a custom Bootstrap build with customised variables. Additionally, since we don't use many of Bootstrap's components, we can reduce the weight of the CSS sent to the browser by only including those components that we do use. Whereas the CDN version includes a 'full-fat' version of Bootstrap with all components included.

Brand colours were taken from the brand guidelines published on the JAC intranet.

---

NB: Enabling SCSS is also useful, as it will be used further down the line for styling form inputs where Bootstrap's default styling doesn't get us far enough.

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,6 +9,7 @@
     "test": "vue-cli-service test:unit"
   },
   "dependencies": {
+    "bootstrap": "^4.3.0",
     "clone": "^2.1.2",
     "firebase": "^5.7.2",
     "firebaseui": "^3.5.1",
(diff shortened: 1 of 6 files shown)


package.json
```diff
@@ -9,6 +9,7 @@
     "test": "vue-cli-service test:unit"
   },
   "dependencies": {
+    "bootstrap": "^4.3.0",
```
tatyree: 4.3.1 is the most current.

ollietreend: @tatyree updated to 4.3.1

PR #12, opened 2019-02-22
Date input component

ollietreend:
This PR introduces a new `DateInput` component for users to input dates.

This component follows GDS guidance on asking users for dates, [as laid out in the GOV.UK Design System](https://design-system.service.gov.uk/patterns/dates/). This means that we ask for dates as 3 input fields: day, month and year.

For example, it is used on the Personal Details page to ask for the user's Date of Birth.

Date of Birth input fields

The component can be bound to data models using `v-model` binding. It will return a JavaScript `Date` object for the user's given input date.

### Example usage

Use `DateInput` component in your Vue component `<template>` block:

```vue
<fieldset>
  <legend>What is your date of birth?</legend>
  <DateInput v-model="date_of_birth" />
</fieldset>
```

In your data model, `date_of_birth` will be a JavaScript `Date` object reflecting the user input.

### DIFF
--- a/tests/unit/components/DateInput.spec.js
+++ b/tests/unit/components/DateInput.spec.js
@@ -1,4 +1,4 @@
-import {shallowMount} from "@vue/test-utils";
+import {shallowMount} from '@vue/test-utils';
 import DateInput from '@/components/DateInput';
 
 describe('components/DateInput', () => {
--- a/tests/unit/components/DateInput.spec.js
+++ b/tests/unit/components/DateInput.spec.js
@@ -1,0 +1,508 @@
+import {shallowMount} from "@vue/test-utils";
+import DateInput from '@/components/DateInput';
+
+describe('components/DateInput', () => {
+  const createTestSubject = (value) => {
+    return shallowMount(DateInput, {
+      propsData: {value}
+    });
+  };
+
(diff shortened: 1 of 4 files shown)


tests/unit/components/DateInput.spec.js
```diff
@@ -0,0 +1,508 @@
+import {shallowMount} from "@vue/test-utils";
```
tatyree: You've got some inconsistent quoting in here.  Sometimes you use single quotes and other times you use double quotes.  I'd suggest we standardise on double, and use single where we need to nest double.

ollietreend: Fair point. That's my inner PHP developer typing single quotes (as is generally preferred in the PHP world), and my IDE inserting double quotes on auto-completed snippets. The main exception being where I've deliberately used double quotes to avoid escaping apostrophes in strings.

I could standardise on double quotes in this file, however the same inconsistency already exists throughout the rest of the codebase.

If this is something we care about standardising on, I'd suggest we accept this as tech debt for now and put a card in the backlog to agree on some style preferences and tighten up our linting rules accordingly. The linting rules are currently loose enough to not care about quote styles.

PR #22, opened 2019-03-01
Add a new `BooleanInput` component

### LINKED ISSUE #21: Boolean Input custom component
A custom component for 'yes/no' inputs.

This follows the GDS pattern of using [inline radios](https://design-system.service.gov.uk/components/radios/#inline-radios) when there are only two options.

The component should bind to data models using `v-model` and return a JavaScript boolean `true` or `false` value depending on the option selected.

### Example use case

Use this component where you're asking a yes/no question.

For example:

Do you have any criminal convictions?

Yes
No

### Done when

Custom component `BooleanInput` exists, and has unit tests.

ollietreend:
Closes #21
Connect #21

This component repeats the common 'yes/no' radio input pattern which is used throughout the form. It's typically used in conjunction with a question, for example:

Do you need any reasonable adjustments?
[radio inputs]
○ Yes    ○ No

Bound to a data model with `v-model`, it will return a JavaScript boolean `true` or `false` given a 'Yes' or 'No' answer respectively.

### Example usage

```vue
<template>
  <fieldset>
    <legend>Do you have any criminal convictions?</legend>
    <BooleanInput v-model="has_criminal_conviction" />
  </fieldset>
</template>
```

---

**Note:** This implements the component discussed in #21, but does not add unit tests so that ticket will not be 'done' after this PR is merged. Unit tests on this component will follow later – this PR is to aid rapidly implementing our end-to-end journey.

### DIFF
--- a/src/components/BooleanInput.vue
+++ b/src/components/BooleanInput.vue
@@ -1,0 +1,44 @@
+<template>
+  <div>
+    <div class="custom-control custom-radio custom-control-inline">
+      <input class="custom-control-input" type="radio" :id="yesInputId" :value="true" v-model="inputValue">
+      <label class="custom-control-label" :for="yesInputId">Yes</label>
+    </div>
+    <div class="custom-control custom-radio custom-control-inline">
+      <input class="custom-control-input" type="radio" :id="noInputId" :value="false" v-model="inputValue">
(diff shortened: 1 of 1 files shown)


tatyree:
I'm adding a 'closes' keyword to this as the described unit tests should be expressed under their own ticket/milestone.  Adding that now.

PR #70, opened 2019-03-14
Email verification

### LINKED ISSUE #16: Stop applicants progressing until they validate their email address
Depends on https://github.com/JudicialAppointmentsCommission/digital-platform/issues/95

## Background

As the email validation is now working (#11), we can now require applicants to validate their emails before they continue with their application. 

## Done when 

- [ ] A user cannot progress their application until they have validated their email.
- [ ] A mechanism exists that allows them to ask the system to resend their validation email.

### LINKED ISSUE #64: Content for validation message on page
When a user logs in who hasn't yet verified their email address, we'll show this message. They can't continue to start their application until they've completed this step.

Relevant guidance from GDS: [Designing the ‘activate your account’ page](https://design-system.service.gov.uk/patterns/confirm-an-email-address/#designing-the-activate-your-account-page)

---

### Verify your email address

We need to make sure we've got the right email address for you.

We've sent an email to **user@example.com** containing a link to verify your email address.

Click the link in the email to complete your registration and start your application.

▼ **Help! I haven't received the email yet**

> It could take up to 10 minutes for our email to arrive.
>
> Don't forget to check your 'spam' or 'junk mail' folder in case it didn't go into your inbox.
>
> If it still hasn't arrived after 10 minutes, you can resend the email:
>
> [ Resend email ]

ollietreend:
Closes #16 
Closes #64

This PR adds a 'Verify your email address' page which is displayed to logged in users who've not yet verified their email address. Users cannot see access any other authenticated route in the application until they've verified their email address.

The page cannot be seen if you've already verified your email address – instead you'll be redirected to the root `/` page.

The verify page contains a 'Resend email' button allowing users to re-send the email. This triggers an `onCall` Firebase Cloud Function.

### Registration happy path

As of this PR, the user registration flow is as follows:

1. `[client]` New user completes the registration form
2. `[server]` Cloud Function `sendVerificationEmailOnNewUser` is triggered when the new user is created
3. `[client]` Vue app displays the 'Verify your email address' page because in the user record `emailVerified === false`
4. User clicks link in email, Firebase updates user record. 'Continue' button takes user back to the application.
5. `[client]` Vue app loads again, this time `emailVerified === true` in the user record, so the user can progress to the application form

### Firebase configuration changes

Whilst working on this PR, I also configured the Firebase project to send use our custom domain in emails.

Emails sent from Firebase (currently 'reset password' and 'email address change') are now sent from the preferred outgoing email at our custom domain.

Action links in emails (including email verification link and reset password link) will use the production domain.

To configure this:

1. In the Firebase console, go to Authentication > Templates
2. Edit any template
3. In the sender/from configuration, click 'customise domain'
4. Under the action URL, click 'customise action URL'

### DIFF
--- a/functions/index.js
+++ b/functions/index.js
@@ -21,13 +21,14 @@ const sendEmail = (email, templateId, personalisation) => {
       {personalisation}
     )
     .then(notifyResponse => {
-      console.info(JSON.parse(JSON.stringify(notifyResponse)));
+      console.info(notifyResponse.body);
       return true;
     });
 };
 
 const sendVerificationEmail = (email) => {
-  return admin.auth().generateEmailVerificationLink(email)
+  const returnUrl = 'https://apply.judicialappointments.digital';
+  return admin.auth().generateEmailVerificationLink(email, {url: returnUrl})
     .then(link => {
       const templateId = functions.config().notify.templates.verification;
       const personalisation = {
--- a/functions/index.js
+++ b/functions/index.js
@@ -1,25 +1,48 @@
 const admin = require("firebase-admin");
-const NotifyClient = require("notifications-node-client").NotifyClient
-
-admin.initializeApp()
 const functions = require("firebase-functions");
+const NotifyClient = require("notifications-node-client").NotifyClient;
 
-exports.sendVerificationEmail = functions.auth.user().onCreate((user) => {
-  const notifyClient = new NotifyClient(functions.config().notify.key)
-  const email = user.email
+admin.initializeApp();
 
-  const sendEmail = (link) => {
-    notifyClient
-      .sendEmail(
-        functions.config().notify.templates.verification,
-        email,
-        { personalisation: { verificationLink: link } }
-      )
-      .then(() => console.info({ type: "verification link", email: email }))
-      .catch(error => console.error(error))
-  }
+const sendEmail = (email, templateId, personalisation) => {
+  const client = new NotifyClient(functions.config().notify.key);
 
+  console.info({
+    action: 'Sending email',
+    email,
+    templateId,
+    personalisation,
+  });
+
+  return client
+    .sendEmail(
+      templateId,
+      email,
+      {personalisation}
+    )
+    .then(notifyResponse => {
+      console.info(JSON.parse(JSON.stringify(notifyResponse)));
+      return true;
+    });
+};
+
+const sendVerificationEmail = (email) => {
   return admin.auth().generateEmailVerificationLink(email)
-    .then(link => sendEmail(link))
-    .catch(error => console.error(error))
+    .then(link => {
+      const templateId = functions.config().notify.templates.verification;
+      const personalisation = {
+        verificationLink: link,
+      };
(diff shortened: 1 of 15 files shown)


functions/index.js
```diff
@@ -1,25 +1,49 @@
...
-        { personalisation: { verificationLink: link } }
-      )
-      .then(() => console.info({ type: "verification link", email: email }))
-      .catch(error => console.error(error))
-  }
+  return client
+    .sendEmail(
+      templateId,
+      email,
+      {personalisation}
```
tatyree: This works?  The docs seem to suggest that it should have a root key of `personalisation:`.

functions/index.js
```diff
@@ -1,25 +1,49 @@
...
+    )
+    .then(notifyResponse => {
+      console.info(notifyResponse.body);
+      return true;
+    });
+};
+
+const sendVerificationEmail = (email) => {
+  const returnUrl = 'https://apply.judicialappointments.digital';
+  return admin.auth().generateEmailVerificationLink(email, {url: returnUrl})
```
tatyree: You might consider rewriting this like:

```
  return admin.auth().generateEmailVerificationLink(email, {url: returnUrl})
    .then((link) => { 
      return link 
     }) \\ or the shorthand `link => link`. 
    .then((link) => {
      const templateId = functions.config().notify.templates.verification;
      const personalisation = {
        verificationLink: link,
      };
      return sendEmail(email, templateId, personalisation);
    });
```

I appreciate it isn't much different, but it makes it clearer that these are separate, logical promises/steps.

functions/index.js
ollietreend: Yep – I was as surprised as you are when reading the Notify docs.

I've gone with the parameters from your original implementation, which seems to work fine. Note whilst the code has been refactored, that particular method call is functionally the same.

The docs also state that a top level parameter `reference` is required, must be unique and must not be `null`. Yet our implementation works without it.

Disappointingly, I think it's an issue with the Notify docs not matching the client library. When you look at the node client implementation, it expects different parameters than those laid out in the docs.

I can show you what I mean tomorrow – afk now.

ollietreend: @tatyree Looks like it's a change which didn't make it into the documentation.

[Documentation](https://docs.notifications.service.gov.uk/node.html#send-an-email-arguments) says pass everything in as top-level parameters

[Actual implementation](https://github.com/alphagov/notifications-node-client/blob/068c6ec65918f6c5206a32bd8b478fe1e1b8a5d1/client/notification.js#L143) says to pass in an `options` param

[Changelog entry](https://github.com/alphagov/notifications-node-client/blob/068c6ec65918f6c5206a32bd8b478fe1e1b8a5d1/CHANGELOG.md#changed-7) suggests this changed in 2017. So not sure why the docs are out of date.

functions/index.js
ollietreend: @tatyree Have a look at my latest changes. Any better? I think the use of `async/await` has improved the readability of this method.

PR #73, opened 2019-03-15
Refactored summary

tatyree:
Intermediate PR - Added Assessors and Character.

### DIFF
--- a/src/components/Review/CharacterTable.vue
+++ b/src/components/Review/CharacterTable.vue
@@ -1,0 +1,65 @@
+<template>
+  <article>
+    <table class="table" v-if="answeredYes">
+      <thead>
+        <tr>
+          <th colspan="10">
+            {{ title }}
+          </th>
+        </tr>
+      </thead>
+      <tbody v-for="record in visibleRecords" :key="record.id">
+        <tr v-for="(value, key) in record" :key="value.id">
+          <th class="text-capitalize offence_key">{{ key }}</th>
+          <td v-if="value instanceof Date" class="offence_info">{{ value.toLocaleDateString("en-GB", { month: "numeric", year: "numeric" }) }}</td>
+          <td v-else class="offence_info">{{ value }}</td>
+        </tr>
+        <tr>
+          <td colspan="10">
+            <RouterLink to="/apply/character" class="float-right">Change</RouterLink>
+          </td>
+        </tr>
+      </tbody>
+    </table>
+
+    <table class="table" v-else>
+      <tbody>
+        <tr>
+          <th class="boolean_no">
+            {{ title }}
+          </th>
+          <td class="p-3 text-right">
+            No
+          </td>
+          <td class="p-3 text-right">
+            <RouterLink to="/apply/character" class="float-right">Change</RouterLink>
+          </td>
+        </tr>
+      </tbody>
+    </table>
+  </article>
+</template>
+
+<script>
+export default {
+  props: {
+    answeredYes: Boolean,
+    records: Object,
+    title: String
+  },
+  computed: {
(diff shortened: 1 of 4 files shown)


src/components/Review/CharacterTable.vue
```diff
@@ -0,0 +1,50 @@
+<template>
+  <table class="table">
+    <thead v-if="answered_yes">
+      <tr>
+        <th colspan="10">
+          {{ title }}
+        </th>
+      </tr>
+    </thead>
+    <tbody v-if="answered_yes" v-for="record in records" :key="record.id">
```
ollietreend: As warned by the linter, `v-for` and `v-if` shouldn't be used on the same element. [The style guide](https://vuejs.org/v2/style-guide/#Avoid-v-if-with-v-for-essential) explains the rationale behind this.

In this case, I would probably split it out into two `<table>` elements which conditionally show if `answered_yes` is true or not, rather than reusing the same table element as you do now.

This will mean you'll likely need to wrap everything in a `<div>`, since there can only be one top-level element inside a `<template>`.

src/views/Sections/Character/Show.vue
```diff
@@ -0,0 +1,143 @@
...
+      <h3 class="card-title">Your Character</h3>
+    </div>
+    <div class="card-body">
+      <h4>Motoring Offences</h4>
+      <CharacterTable
+        title="Have you ever been convicted of, cautioned for or received a
+        conditional caution for a motoring offence (other than a parking
+        offence)?"
+        v-bind:records="applicant.character_motoring_offences"
+        v-bind:answered_yes="applicant.character_has_motoring_offences"
```
ollietreend: When binding values to attributes/properties, I tend to use the [Vue shorthand](https://vuejs.org/v2/guide/syntax.html#Shorthands). So in this case, I'd suggest rewriting `v-bind:records` and `v-bind:answered_yes` as `:records` and `:answered_yes`, respectively.

I'd also suggest changing the `answeres_yes` property to be camel-cased. We've not used underscored property names elsewhere in the project (I didn't actually think you could since they're not standard in HTML).
The docs do talk about the difference between [camelCased vs kebab-cased properties](https://vuejs.org/v2/guide/components-props.html#Prop-Casing-camelCase-vs-kebab-case), but since we're not rendering in-DOM templates, the case sensitivity doesn't affect us.

src/components/Review/CharacterTable.vue
```diff
@@ -0,0 +1,59 @@
...
+
+<script>
+export default {
+  props: {
+    answered_yes: Boolean,
+    records: Object,
+    title: String
+  },
+  computed: {
+    visibleRecords: function () {
```
ollietreend: (I wrote this before you fixed the linter error, but it's somewhat still relevant)

As warned by the linter, `v-for` and `v-if` shouldn't be used on the same element. [The style guide](https://vuejs.org/v2/style-guide/#Avoid-v-if-with-v-for-essential) explains the rationale behind this.

In this case, I would probably split it out into two `<table>` elements which conditionally show if `answered_yes` is true or not, rather than reusing the same table element as you do now.

This will mean you'll likely need to wrap everything in a `<div>`, since there can only be one top-level element inside a `<template>`.

### Update

I think the readability of the branching logic might be improved by having two `<table>` elements in this component, rather than 'turning off' the `<tbody>` by returning an empty object in the computed property.

tatyree: Yeah, two tables were definitely the way to go.

tatyree: Oh, returning the empty object in the computed property is just there to keep the linter happy.   I wasn't checking for it at the time of your comment.

PR #78, opened 2019-03-16
Self assessment snagging

ollietreend:
To be reviewed by @ClaireTroughtonJAC in conjunction with my comments on JudicialAppointmentsCommission/digital-platform#144

If time permits, I think there are some tweaks to be made before this is ready for PR.

---

Closes JudicialAppointmentsCommission/digital-platform#144

### DIFF
--- a/src/views/Sections/SelfAssessment/Edit.vue
+++ b/src/views/Sections/SelfAssessment/Edit.vue
@@ -16,10 +16,57 @@
       </div>
 
       <fieldset :disabled="application.state === 'submitted'">
-        <legend>Additional Selection Criteria</legend>
+        <legend>Self Assessment</legend>
+        <div class="fieldset-text">
+          <p>Please refer to the
+          <a href="https://www.judicialappointments.gov.uk/sites/default/files/sync/128_final_competency_framework.pdf" target="_blank">guidance and competency framework</a>
+            on the JAC website before completing this part of your application.</p>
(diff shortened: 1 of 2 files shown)


src/views/Sections/SelfAssessment/Edit.vue
```diff
@@ -4,46 +4,66 @@
...
+          <label for="additional_selection_criteria_details">Please provide details</label>
+          <textarea class="form-control" id="additional_selection_criteria_details" rows="9" v-model="application.additional_selection_criteria_details"></textarea>
+        </div>
+      </fieldset>
+
+      <fieldset>
+        <legend>Reasonable Length of Service</legend>
+        <div class="fieldset-text">Refer to the Information Page</div>
+        <div class="form-group">
+          <label for="length_of_service_mitigation">If you are unable to provide 3 years’ service, please provide any mitigation you would like the Commission to consider</label>
```
tatyree: No need to do this for this commit, but could you set your wrap a bit narrower and insert line breaks between tags and content, please? 

For example:
```
<label for="length_of_service_mitigation">
  If you are unable to provide 3 years’ service, please provide any mitigation
  you would like the Commission to consider
</label>
```
That one is wrapped at 80, but I'd be just as happy with 130.

ollietreend: I've added this comment to our linting card to pick up later.
JudicialAppointmentsCommission/digital-platform#87

PR #139, opened 2019-04-16
Enable function to create/update QT summary records based on submissions.

tatyree:
closes https://trello.com/c/OPlBbemx/15-as-a-jac-admin-i-want-each-test-phase-to-be-timed-on-submission

This includes the following commit: 

```
Reorganise QT functions

This is an intermediate step towards something like this:

https://codeburst.io/organizing-your-firebase-cloud-functions-67dc17b3b0da

I considered doing a full refactor to achieve an automated layout as
described in this article, but ultimately decided that doing so was
premature optimisation. I will probably reconsider this as I add more
functions, so this should not be taken to be the final layout of our
functions.
```

### DIFF
--- a/functions/qt/submission.js
+++ b/functions/qt/submission.js
@@ -43,7 +43,7 @@ const createRecord = async (record) => {
 }
 
 exports = module.exports = functions.https.onRequest((request, response) => {
-  createRecord(request.body)
+  return createRecord(request.body)
     .then(() => {
       return response.status(200).send({status: 'OK'});
     })
--- a/functions/qt/submission.js
+++ b/functions/qt/submission.js
@@ -8,8 +8,8 @@ const createRecord = async (record) => {
   console.info({ createRecordCalled: record });
 
   // This function expects the test title to be in the following format:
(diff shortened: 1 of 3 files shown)


functions/qt/submission.js
```diff
@@ -41,3 +42,14 @@ exports.createRecord = async (record) => {
   return true;
 }
 
+exports = module.exports = functions.https.onRequest((request, response) => {
```
ollietreend: The first `exports =` is unnecessary here. Simply using `module.exports = functions....` will do what you want.

functions/qt/summary.js
```diff
@@ -0,0 +1,23 @@
+const admin = require("firebase-admin");
+const functions = require("firebase-functions");
+
+exports = module.exports = functions.firestore.document("qtSubmissions/{submission}").onCreate((snapshot, context) => {
```
ollietreend: Ditto here, re unnecessary use of `exports = `

functions/qt/summary.js
```diff
@@ -0,0 +1,23 @@
...
+      {
+        finishedAt: data.createdAt
+      }
+    }
+  };
+
+  return firestore
+    .collection("qtSummaries")
+    .doc(data.userId)
+    .set(record)
```
ollietreend: Use of `set` will overwrite existing record data.

PR #140, opened 2019-05-13
Format exported application dates as YYYY-MM-DD

ollietreend:
This commit changes the script `functions/exportSubmitted.js`, which is used to produce a JSON file containing data of submitted applications, so that dates are formatted as YYYY-MM-DD.

To do this, it monkey patches the `toJSON` method on `Date` objects so that the `JSON.stringify()` method (which is used to produce the JSON export file) renders them in the desired format.

This is a hacky approach, but it works as intended in this context and was done under time pressure. It should be replaced with something more sensible in the future.

### DIFF
--- a/functions/exportSubmitted.js
+++ b/functions/exportSubmitted.js
@@ -68,6 +68,24 @@ const getRecords = async () => {
   return records;
 };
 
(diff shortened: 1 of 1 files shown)


ollietreend:
Closing – this feature is no longer required

PR #155, opened 2019-06-18
Convert AoA answers to objects before sending

tatyree:
See comments for a full explaination. Currently this needs to be
manually attached to the Google Form Apps script handling the
submission. Instructions are in the RUCA repo, but I will move them over
to this repo shortly.


google-apps-scripts/google-forms/writeAnswersToFirestore.js
```diff
@@ -0,0 +1,69 @@
...
+
+         Where the populated element indicates the row position of the user's chosen answer. It is an array
+         because the checkbox grid needs to be able to deal with cases where there may be multiple answer columns chosen.
+         We DO NOT support this format--anything other than two columns will cause problems. That said, these two columns can be
+         named any way the ops teams/policy like.
+         */
+      checkboxGrid.getColumns().forEach(function (column) {
+        column = column.toString();
+        gridAnswers[column] = rows[answersArray.indexOf(column)];
+      });
```
ollietreend: As it stands, this will produce an object with the following format:

```json
{
  "The answer selected by the user will be here": "Most appropriate",
  "And these answers are usually quite lengthy": "Least appropriate"
}
```

I propose that flipping the key & value in the object would produce a more sensible / useful output for storage and post-processing:

```json
{
  "Most appropriate": "The answer selected by the user will be here",
  "Least appropriate": "And these answers are usually quite lengthy"
}
```

That way, a score could be calculated by simply inspecting the value of `response['Least appropriate']` and `response['Most appropriate']`. This would likely also be the way we'd want to display it in a user-facing view of answers given.

I realise that either format would be usable, but the latter seems like the more simple / straightforward way to store the results given our likely read patterns.

What do you think @tatyree?

google-apps-scripts/google-forms/writeAnswersToFirestore.js
```diff
@@ -0,0 +1,69 @@
...
+         [['Most Appropriate'],null,null,null,['Least Appropriate'],null]
+
+         Where the populated element indicates the row position of the user's chosen answer. It is an array
+         because the checkbox grid needs to be able to deal with cases where there may be multiple answer columns chosen.
+         We DO NOT support this format--anything other than two columns will cause problems. That said, these two columns can be
+         named any way the ops teams/policy like.
+         */
+      checkboxGrid.getColumns().forEach(function (column) {
+        column = column.toString();
+        gridAnswers[column] = rows[answersArray.indexOf(column)];
```
ollietreend: To really make it obvious exactly what we're doing here, could we move the `rows[answersArray.indexOf(column)]` into its own variable and then perform the assignment in a separate line?

So you'd end up with:

```js
column = column.toString();
const row = rows[answersArray.indexOf(column)];
gridAnswers[column] = row;
```

google-apps-scripts/google-forms/writeAnswersToFirestore.js
tatyree: I think you've misread that @ollietreend.  Here is an example of the actual output:

```
[19-06-20 14:29:57:511 BST] {
  "title": "Part 1: Array of Arrays Issue",
  "email": "todd.tyree@judicialappointments.digital",
  "Question 1: You have before you a case ": {
    "Most Appropriate": "A. Gain ",
    "Least Appropriate": "C. Adjourn"
  },
  "Question 2: Lorem Ipsum": {
    "Most Appropriate": "A. dolor hoc",
    "Least Appropriate": "B. ad hoc"
  }
}
```

ollietreend: Ah, okay! So it was that way around all along. My mistake.

google-apps-scripts/google-forms/writeAnswersToFirestore.js
tatyree: @ollietreend - I've changed a couple of the variable names and done a bit of refactoring to make the intention clearer.

google-apps-scripts/google-forms/writeAnswersToFirestore.js
```diff
@@ -0,0 +1,70 @@
...
+
+         Where the populated element indicates the row position of the user's chosen answer. It is an array
+         because the checkbox grid needs to be able to deal with cases where there may be multiple answer columns chosen.
+         We DO NOT support this format--anything other than two columns will cause problems. That said, these two columns can be
+         named any way the ops teams/policy like.
+         */
+      checkboxGrid.getColumns().forEach(function (columnName) {
+        columnName = columnName.toString();
+        const chosenAnswer = allRows[answersArray.indexOf(columnName)];
+        userSelectedAnswers[column] = chosenAnswer;
```
ollietreend: `userSelectedAnswers[column]` should be `userSelectedAnswers[columnName]` to match the new variable name

tatyree: Good catch, thanks!

PR #163, opened 2019-06-25
Process form submissions

tatyree:
This PR finalises what we hope to be the last iteration for a while on the Test submission processor https endpoint script.


functions/qt/submission.js
```diff
@@ -2,38 +2,49 @@ const admin = require("firebase-admin");
...
+
+  /*
+   * We only record the `finishedAt` time for the *first* submission. In the event of duplicate submissions, we save the each
+   * additional one and count the number of duplicates in the `usersTests` record. We do not update the `finishedAt`
+   * timestamp again and we do not send duplicate confirmation emails.
+   *
+   */
+  const userTestDoc = await userTest.get();
+  if (userTestDoc.exists) {
+    const data = userTestDoc.data()
```
ollietreend: Please end the line with a `;` semicolon

functions/qt/submission.js
```diff
@@ -2,38 +2,49 @@ const admin = require("firebase-admin");
...
+  const userTest = firestore
+    .collection("usersTests")
+    .doc(record["Unique reference code - for JAC use only"]);
+
+  record.userTest = userTest;
 
-  record.createdAt = new Date();
-  await firestore
-    .collection('qtSubmissions')
+  const uts = firestore
```
ollietreend: `userTestSubmission` could be a more meaningful name for the variable `uts`

functions/qt/submission.js
```diff
@@ -2,38 +2,49 @@ const admin = require("firebase-admin");
...
+  console.info({ createdUserTestSubmission: uts.id });
+
+  /*
+   * We only record the `finishedAt` time for the *first* submission. In the event of duplicate submissions, we save the each
+   * additional one and count the number of duplicates in the `usersTests` record. We do not update the `finishedAt`
+   * timestamp again and we do not send duplicate confirmation emails.
+   *
+   */
+  const userTestDoc = await userTest.get();
+  if (userTestDoc.exists) {
```
ollietreend: It could be useful to also include an `else` statement to capture form submissions where the `userTest` document *doesn't* exist. This would likely mean that the submitted 'unique reference code' field was changed to an unexpected value which doesn't exist in the database.

We could log the field value in this case so we can keep track of any mismatched submissions.

functions/qt/submission.js
```diff
@@ -2,38 +2,49 @@ const admin = require("firebase-admin");
...
+      console.info({updatedUsersTests: userTest.id});
+
+      const user = await admin.auth().getUser(data.userUid);
+      const client = new NotifyClient(functions.config().notify.key);
+      const personalisation = {firstname: user.displayName};
+      await client.sendEmail(record.confirmationTemplate, user.email, {personalisation});
+      console.info({sentConfirmationEmail: user.email});
+    } else {
+      const increment = admin.firestore.FieldValue.increment(1);
+      userTest.update({duplicateSubmissions: increment});
```
ollietreend: Is `duplicateSubmissions` a useful piece of data for us to store in the database? What's the intended use case for this field?

If we don't want/need to keep it long term, it might be better to log it instead.

That way, we'll be able to query against the logs from a test day to uncover any unexpected behaviour. Similar to my other comment about logging submissions where the unique reference code doesn't match up with a `userTest` document. Both duplicate submissions and unexpected submissions should hopefully be edge cases, but inspecting logs would help us uncover the cases where they do occur.

functions/qt/submission.js
```diff
@@ -2,38 +2,49 @@ const admin = require("firebase-admin");
 const functions = require("firebase-functions");
 const NotifyClient = require("notifications-node-client").NotifyClient;
 
-// This throws exceptions rather than logging because we want to know as soon as these issues occur.  By throwing, we are
-// ensuring they will appear in StackDriver Errors in an actionable way.
 const createRecord = async (record) => {
```
ollietreend: I can see that the `createRecord` function has 3 responsibilities:

1. It creates a `userTestSubmissions` document
2. It updates a `usersTests` document
3. It sends a confirmation email to the user

I think it could be easier to follow if these were split into their own functions, with a 'main' function tying them all together.

Additionally, could we rename `createRecord` to something more meaningful and descriptive of what the function actually does – perhaps `processFormSubmission`?

functions/qt/submission.js
```diff
@@ -2,43 +2,72 @@ const admin = require("firebase-admin");
...
-// This throws exceptions rather than logging because we want to know as soon as these issues occur.  By throwing, we are
-// ensuring they will appear in StackDriver Errors in an actionable way.
-const createRecord = async (record) => {
-  const firestore = admin.firestore();
-  const client = new NotifyClient(functions.config().notify.key);
+const createUserTestSubmission = async (record, firestore) => {
+  const finishedAt = admin.firestore.FieldValue.serverTimestamp();
+  record.finshedAt = finishedAt;
+
+  const userTest = firestore
```
ollietreend: @tatyree I'd be tempted to move the `const userTest = ` up into the top-level `main` function. The `userTest` document is a dependency of all 3 'sub' functions, but right now that's not entirely clear because it's defined in `createUserTestSubmission` and then passed back as a return value and into the other functions.

ollietreend: Maybe even perform the `.get()` in the top-level `main`, too.

```
const userTest = firestore.collection(...).doc(...);
const userTestSnapshot = userTest.get();
```

Both are needed throughout the other 3 functions, I think..

PR #172, opened 2019-07-04
Move questions to Outreach page and add Show component for Outreach

alexbokii:
**Changelog:**

Trello ticket 42

1) Move questions:
- Have you attended an outreach event on JAC selection exercises?
- Did you participate in a Judicial Work Shadowing Scheme?
from Diversity section to Outreach

2) Add Show component for Outreach as it was missing

3) Update text in the legend for 'How did you hear about this vacancy?' question


hosting/apply/src/views/Sections/Outreach/Edit.vue
```diff
@@ -1,10 +1,10 @@
 <template>
   <section>
     <form @submit.prevent="save">
-      <h2>How did you hear about this vacancy?</h2>
 
       <fieldset>
-        <legend>Select any that apply:</legend>
+        <legend>How did you hear about this vacancy?</legend>
+        <div class="fieldset-text">Select any that apply:</div>
```
ollietreend: For consistency with other pages of the form, can we please change this to read:

> Select all that apply

including dropping the `:` colon at the end

hosting/apply/src/views/Sections/Outreach/Edit.vue
```diff
@@ -15,6 +15,16 @@
...
       </fieldset>
 
+      <fieldset>
+        <legend>Have you attended an outreach event on JAC selection exercises?</legend>
+        <SelectList id="has_attended_outreach_event" :options="selectListOptions.yesNo" v-model="applicant.has_attended_outreach_event" />
+      </fieldset>
+
+      <fieldset>
+        <legend>Did you participate in a Judicial Work Shadowing Scheme?</legend>
+        <SelectList id="participated_in_judicial_work_shadowing_scheme" :options="selectListOptions.yesNo" v-model="applicant.participated_in_judicial_work_shadowing_scheme" />
```
ollietreend: Rather than using a `SelectList` component here, paired with the options in `selectListOptions.yesNo`, I think we should be able to simply use a `BooleanInput` component instead.

The `BooleanInput` component provides 'yes/no' radio buttons and results in a `true`/`false` value being set in the data model. Whereas the `SelectList` is more generic and will in this case set the values to `"Yes"`/`"No"`.

The reason we've used `SelectList` on the Diversity page is because we want people to be able to opt-out of answering the question – i.e. by providing a response of 'Prefer not to answer'. In this case, for these two questions, I don't think that's necessary to 'yes/no' should suffice.

hosting/apply/src/views/Sections/Outreach/Show.vue
```diff
@@ -0,0 +1,43 @@
...
+    <div class="card-body">
+
+      <ShowSelectList
+        changeLink="/apply/outreach"
+        :hasOther="applicant.how_did_you_hear_has_other"
+        title="How did you hear about this vacancy?"
+        :records="applicant.how_did_you_hear"
+        />
+
+      <ShowSelectList
```
ollietreend: Changing the two moved questions from `SelectList` to `BooleanInput` inputs will mean this 'show' page needs to change how it displays the fields, since the values will change from `"Yes"`/`"No"` to `true`/`false`.

I don't think there's a component for displaying the value of `BooleanInput`s, because elsewhere they've been paired with other fields in the 'show' views – mostly `BooleanWithCheckboxes`, which doesn't really apply here.

The simplest way of displaying these fields manually. Using `views/Sections/Citizenship/Show.vue` as a basis, you could use something like:

```vue
<table class="table">
  <tbody>
    <tr>
      <th scope="row">Have you attended an outreach event on JAC selection exercises?</th>
      <td>{{ applicant.has_attended_outreach_event ? 'Yes' : 'No' }}</td>
      <td>
        <RouterLink to="/apply/outreach">Change</RouterLink>
      </td>
    </tr>
  </tbody>
</table>
```

hosting/apply/src/views/Sections/Outreach/Show.vue
```diff
@@ -0,0 +1,43 @@
+<template>
```
ollietreend: Since you've added another 'show' component which will be displayed be displayed below the others on the 'review' page, please add the class `mb-3` to the `Diversity/Show` component.

This adds margin to the bottom of the Diversity Show component.

ollietreend: @alexbokii Sorry – this wasn't very clear. But could you put the `mb-3` in `Diversity/Show.vue` *not* `Outreach/Show.vue`?

I couldn't comment in the other file because it hasn't changed in this PR.

But that should solve the margin spacing issue.

ollietreend: This is what I'm referring to when I talk about the spacing: every other section on the review page has a bottom margin so it doesn't touch the next section. But diversity was always the final section, so didn't have a bottom margin. Now we need to add one because it's not the last section any more.

Currently it looks like this:

Application_Form_-_Judical_Appointments_Commission_no_spacing

But we want margin bottom like this, using `mb-3` on `Diversity/Show.vue`:

Application_Form_-_Judical_Appointments_Commission_spacing

alexbokii: Thanks, @ollietreend, it's done now!

PR #182, opened 2019-07-05
Remove "We’re looking for people" section

alexbokii:
**Changelog:**

Remove "We’re looking for people" section from Review and Submit page

Trello card 175

### DIFF
--- a/hosting/apply/src/views/Sections/Review/Edit.vue
+++ b/hosting/apply/src/views/Sections/Review/Edit.vue
@@ -12,14 +12,12 @@
     <Declarations />
     <Diversity />
     <Outreach />
-    <form v-if="application.state != 'submitted'"  @submit.prevent="save">
-      <div class="form-actions">
-        <button class="btn btn-primary mr-2" type="button" @click.prevent="saveAndSubmit">
-          Submit application
-        </button>
-        <span class="spinner-border spinner-border-sm text-secondary" v-if="isSaving"></span>
-      </div>
-    </form>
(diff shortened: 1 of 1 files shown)


hosting/apply/src/views/Sections/Review/Edit.vue
```diff
@@ -16,14 +16,6 @@
       You are applying for <strong>{{vacancy.jac_ref}}: {{vacancy.title}}</strong>
     </p>
     <form v-if="application.state != 'submitted'"  @submit.prevent="save">
```
ollietreend: Please remove the `<form>` element – this is not needed now that we're not collecting input data from the user.

hosting/apply/src/views/Sections/Review/Edit.vue
```diff
@@ -16,14 +16,6 @@
...
-      <fieldset>
-        <legend>
-          We’re looking for people to take part in user research activities
-          such as testing a future version of this form. Tell us if you’re
-          happy for us to contact you.
-        </legend>
-        <BooleanInput v-model="application.willing_to_participate_in_user_research" />
-      </fieldset>
       <div class="form-actions">
         <button class="btn btn-primary mr-2" type="button" @click.prevent="saveAndSubmit">
```
ollietreend: The `type="button"` attribute isn't needed here so can be removed. It would have been needed if it was a secondary button in a `<form>` (e.g. when the 'Save' button is secondary to 'Save and continue' in a form).

PR #190, opened 2019-07-15
Refactor QT submission function

alexbokii:
**Change log:**

1) Refactor submission so we set "finishedAt" value before saving the test submission.
2) Add unit tests


functions/qt/submission.test.js
```diff
@@ -0,0 +1,155 @@
...
+const REFERENCE_CODE_FIELD_NAME = 'Unique reference code - for JAC use only';
+
+let mockauth, mockfirestore, mocksdk, firebaseAdmin, emailSent;
+
+beforeEach(async () => {
+  mockauth = new firebasemock.MockAuthentication();
+  mockfirestore = new firebasemock.MockFirestore();
+  // we need only auth() and firestore(), so we return null
+  // for all other services
+  mocksdk = new firebasemock.MockFirebaseSdk(
```
tatyree: Trying to do this by building jest mocks was what caused me to tear my hair out.  This is a huge improvement.

functions/qt/submission.test.js
```diff
@@ -0,0 +1,155 @@
...
+  // as firebase-mock supports `firestore.FieldValue.serverTimestamp` only for `update` operations,
+  // because we use `set` operation to save userTestSubmission, `finishedAt` field won't be populated
+  // this is why we replace `serverTimestamp` implementation with a custom object to
+  // assert on this object to make sure that the same `finishedAt` timestamp is used for both
+  // userTest and userTestSubmission
+  firebaseAdmin.firestore.FieldValue.serverTimestamp = () => ({isEqual: () => false, type: 'submission-test'});
+
+});
+
+function setupEmailMock(f) {
```
tatyree: No need to do this now, but it is probably worth abstracting this out into a helper or even its own module in the near future.

functions/qt/submission.test.js
```diff
@@ -0,0 +1,155 @@
...
+
+  const res = {
+    status: (code) => ({
+      send: (payload) => {
+        testHandler(payload);
+      }
+    })
+  };
+
+  it('should save a new submission and send notification to user', (done) => {
```
tatyree: Kill the 'should', and reword the object please.  I find it helps to read it as a sentence:
Contrast this: 

'Submission should save a new submission and send notification to user' 

with something like:

'Submission saves the answers from Google Forms and sends an email to user' 

Or similar.

functions/qt/submission.test.js
```diff
@@ -0,0 +1,155 @@
...
+      expect(userTestSubmission[REFERENCE_CODE_FIELD_NAME]).toBe('test-reference-code');
+      expect(emailSent).toBe(true);
+      return done();
+    };
+
+    const submission = require('./submission.js');
+    submission(req, res);
+  })
+
+  it('does not save submission or send email when userTest is missing', (done) => {
```
tatyree: Again, we know it's a submission already...

functions/qt/submission.test.js
```diff
@@ -0,0 +1,155 @@
...
+      expect(emailSent).toBe(false);
+
+      return done();
+    };
+
+    const submission = require('./submission.js');
+    submission(req, res);
+  });
+
+  it('should not save timestamp if finishedAt already exists', (done) => {
```
tatyree: ditto.

functions/qt/submission.test.js
alexbokii: 👍 we can do this before adding tests for other functions

PR #200, opened 2019-07-26
QT: add iframe into the new Test page

alexbokii:
**Change log:**

1) Add new page Take Test (the page with iframe) and tests
2) Rename old TakeTest component into Test component
3) Move Loading spinner and error message into a separate component which is used in Take Test and Test
4) Create loadTestData helper which is used in Take Test and Test


hosting/qt/src/components/LoadingMessage.vue
```diff
@@ -0,0 +1,25 @@
...
+    >
+      <p>Could not load data. Please reload the page and try again.</p>
+    </div>
+  </div>
+</template>
+
+<script>
+export default {
+  props: {
+    loadfailed: Boolean,
```
ollietreend: It would be more consistent if this variable name was camel cased – so `loadFailed` rather than `loadfailed`

hosting/qt/src/main.scss
```diff
@@ -50,3 +51,17 @@ h2 {
...
+
+#app {
+  height: 100%;
+  display: flex;
+  flex-direction: column;
+}
+
+main {
+  flex-grow: 1;
+}
```
ollietreend: Please end all files with a newline character.

Also, don't forget to configure your editor to read `.editorconfig` files so that newline file endings are handled automatically 🙂 [Here's the Sublime Text plugin](https://github.com/sindresorhus/editorconfig-sublime)

hosting/qt/src/utils/helpers/loadTestData.js
```diff
@@ -0,0 +1,14 @@
+const loadTestData = async (store, route) => {
```
ollietreend: Let's refactor the entire `utils` directory to be `helpers` and drop the 2nd level nested `helpers` directory.

So `sanitizeFirestore` and `loadTestData` sit at the same level as helpers.

hosting/qt/src/views/TakeTest.vue
```diff
@@ -1,115 +1,99 @@
...
       </div>
-
-      <TestCard
-        v-if="!testHasClosed"
-        class="mt-4"
+    </div>
+    <div class="iframe-conainer">
+      <iframe
+        id="inlineFrameExample"
+        title="Inline Frame Example"
```
ollietreend: Are the `id` and `title` attributes on this element needed here? They seem like demo / boilerplate attributes – probably not needed for production?

hosting/qt/src/views/TakeTest.vue
```diff
@@ -1,115 +1,99 @@
...
+        width="100%"
+        height="100%"
+        frameborder="0"
```
ollietreend: The `frameborder` attribute [is deprecated](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/iframe#Deprecated_attributes) and not used in HTML5.

Please remove and replace with a CSS rule `border: 0` on the `<iframe>` element.

hosting/qt/src/views/TakeTest.vue
```diff
@@ -1,115 +1,99 @@
...
-  },
-  methods: {
-    loadTestData() {
-      this.$store.commit('setTestId', this.$route.params.id);
-      return Promise.all([
-        this.$store.dispatch('loadTest'),
-        this.$store.dispatch('loadUserTest'),
-      ])
-        .then(() => {
+    const formUrlExistsInState = this.testFormUrl;
```
ollietreend: Is the variable `formUrlExistsInState` needed here? It might be clearer to just use `this.testFormUrl` directly in the `if()` statement, because it looks like it's not used anywhere else.

hosting/qt/src/utils/helpers/loadTestData.js
```diff
@@ -0,0 +1,14 @@
...
+  store.commit('setTestId', route.params.id);
+  try {
+    await store.dispatch('loadTest');
+    await store.dispatch('loadUserTest');
+    await store.dispatch('subscribeUserTest');
```
ollietreend: Let's talk about this in person – I don't think it does quite what we want it to.

hosting/qt/src/views/TakeTest.vue
```diff
@@ -1,115 +1,99 @@
...
+        }
+      });
+    }
+  },
+  methods: {
+    leaveThePage() {
+      if (confirm('If you leave this page the progress will be lost. Do you want to leave the page?')) {
+        this.$router.go(-1);
+      } else {
+        return;
```
ollietreend: Is the `else {}` block needed here? It doesn't look like it does anything.

hosting/qt/tests/unit/components/LoadingMessage.spec.js
```diff
@@ -0,0 +1,26 @@
+import {shallowMount} from '@vue/test-utils';
+import LoadingMessage from '@/components/LoadingMessage';
+
+describe('components/IEWarning', () => {
```
ollietreend: The top level description should be `'components/LoadingMessage'`. Please change.

hosting/qt/tests/unit/utils/loadTestData.spec.js
```diff
@@ -0,0 +1,38 @@
+import loadTestData from '@/utils/helpers/loadTestData.js';
+
+describe('loadTest', () => {
```
ollietreend: Please name the description the same as the file. So here it'd be `helpers/loadTestData` (assuming it's been moved/refactored to top level `helpers` directory)

hosting/qt/tests/unit/views/TakeTest.spec.js
```diff
@@ -1,253 +1,81 @@
...
-
-        it('commits the hardcoded Qualifying Test ID to the store', () => {
-          const expectQtId = 'ANEMLDHK21g6HtnXQBiC';
-          expect(store.commit).toHaveBeenCalledWith('setQtId', expectQtId);
-        });
-
-        it('executes `this.loadQtData`', () => {
-          expect(mockLoadQtData).toHaveBeenCalled();
-        });
+    it('renders a TakeTest component', () => {
```
ollietreend: This description might be more accurate as: "It renders an iframe"

hosting/qt/tests/unit/views/TakeTest.spec.js
```diff
@@ -1,253 +1,81 @@
...
+      },
+      store = new Vuex.Store({
+        getters,
+      });  
+    });
 
-        it('says the QT has closed', () => {
-          expect(wrapper.text()).toContain('This qualifying test has now closed.');
-        });
+    it('runs loadTestData function', () => {
```
ollietreend: This description seems to be wrong.

PR #206, opened 2019-08-01
Reorganise store.js into namespaced modules

tatyree:
**To start, this is for discussion.  DO NOT MERGE YET**

I've been using this pattern a lot on my personal sites and find it is much easier to keep track of.  I've not included vuexfire yet as the firebase dependency would clash with work @alexbokii is currently doing.  However, I've left the `Promise.all` loader code in `index.js`, as this is how I normally trigger the firestore bindings.  It will be obvious how it works once I put vuexfire in.


hosting/apply-admin/src/store/index.js
```diff
@@ -0,0 +1,26 @@
...
+import applications from '@/store/modules/applications';
+
+Vue.use(Vuex);
+
+const store = new Vuex.Store({
+  strict: process.env.NODE_ENV !== 'production',
+  actions: {
+    // eslint-disable-next-line
+    init: async ({ dispatch }) => {
+      await Promise.all([]);
```
tatyree: It will become obvious why I've left this in once I start binding records with vuexfire.

PR #211, opened 2019-08-06
Vuexfire

tatyree:
This PR creates the first `vuexfire` bound store, `vacancies`.  

This only implements the binding.  The admin pages still need to be built.

Massive thanks to @ollietreend and @alexbokii  for working out how to test the binding using jest.

### DIFF
--- a/hosting/apply-admin/package.json
+++ b/hosting/apply-admin/package.json
@@ -13,7 +13,6 @@
   "dependencies": {
     "core-js": "^2.6.5",
     "firebase": "^6.2.4",
-    "firebase-mock": "^2.2.10",
     "firebaseui": "^4.0.0",
     "govuk-frontend": "^3.0.0",
     "vue": "^2.6.10",
@@ -32,6 +31,7 @@
     "babel-jest": "^23.6.0",
     "eslint": "^5.16.0",
     "eslint-plugin-vue": "^5.0.0",
+    "firebase-mock": "^2.2.10",
     "sass": "^1.17.4",
     "sass-loader": "^7.1.0",
     "vue-template-compiler": "^2.6.10"
--- a/hosting/apply-admin/package.json
+++ b/hosting/apply-admin/package.json
@@ -13,11 +13,13 @@
   "dependencies": {
     "core-js": "^2.6.5",
     "firebase": "^6.2.4",
+    "firebase-mock": "^2.2.10",
     "firebaseui": "^4.0.0",
     "govuk-frontend": "^3.0.0",
     "vue": "^2.6.10",
     "vue-router": "^3.0.3",
-    "vuex": "^3.0.1"
+    "vuex": "^3.0.1",
+    "vuexfire": "^3.1.2"
   },
   "devDependencies": {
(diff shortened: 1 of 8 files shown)


hosting/apply-admin/package.json
```diff
@@ -13,11 +13,13 @@
   "dependencies": {
     "core-js": "^2.6.5",
     "firebase": "^6.2.4",
+    "firebase-mock": "^2.2.10",
```
ollietreend: `firebase-mock` is a dev-only dependency, so should really be in `devDependencies` rather than `dependencies`

hosting/apply-admin/src/store/vacancies.js
```diff
@@ -0,0 +1,14 @@
+import { firestore } from '@/firebase';
+import { firestoreAction } from 'vuexfire';
+
+export default {
+  namespaced: true,
+  actions: {
+    bindVacancies: firestoreAction(context => {
```
ollietreend: Since this is a namespaced module, would it make more sense for the action to simply be called `bind`?

So we'd call `vacancies/bind` rather than `vacancies/bindVacancies`?

It may even be more accurate for this action to be called `bindCollection` – i.e. to bind the entire `vacancies` collection – which would leave room for us to `bind` other refs in this module later.

Just a thought. I'm not wedded to the idea.

hosting/apply-admin/tests/unit/store/vacancies.spec.js
```diff
@@ -0,0 +1,36 @@
+import vacancies from '@/store/vacancies';
+// This is required by `jest.mock('vuexfire...`, but the linter can't interpret the mock.
+// eslint-disable-next-line
+import { firestoreAction } from 'vuexfire';
```
ollietreend: Actually – this isn't required for the `jest.mock` itself, rather it's required if we want to assert anything against that mocked module in this test spec file.

e.g. we were doing something like:

```js
expect(firestoreAction).toHaveBeenCalled()
```

Since we're not doing this any more, we actually don't need to import this module. The `jest.mock()` will still work. We can safely drop this line and the explanatory comment.

package.json
```diff
@@ -6,5 +6,8 @@
   "readme": "README.md",
   "devDependencies": {
     "firebase-tools": "^7.0.2"
+  },
+  "dependencies": {
+    "vuexfire": "^3.1.0"
```
ollietreend: It looks like you've added `vuexfire` as a top-level dependency to the entire repo. This isn't needed, so there shouldn't be any changes to the top-level `package.json` or `package-lock.json`

hosting/apply-admin/src/store.js
```diff
@@ -10,12 +12,22 @@ const store = new Vuex.Store({
...
   state: {},
-  mutations: {},
-  actions: {},
+  mutations: {
+    ...vuexfireMutations,
+  },
+  actions: {
+    init: async ({ dispatch }) => {
+      await Promise.all([
+        dispatch('vacancies/bindVacancies'),
```
ollietreend: Don't forget to change `vacancies/bindVacancies` to `vacancies/bind` here.

PR #213, opened 2019-08-07
Apply-admin: Display a list of vacancies on a dashboard page

alexbokii:
**Change log:**

1) Create a new dashboard component instead of home component and rename route to the 'dashboard' (as per the prototype)
2) Update test for Sign in journey to reflect the renaming of the dashboard route, remove firebase mock from the test as it's not used


hosting/apply-admin/src/router.js
```diff
@@ -3,7 +3,7 @@ import Router from 'vue-router';
 import store from '@/store';
 
 // Views
-import Home from './views/Home.vue';
+import Dashboard from './views/Dashboard.vue';
```
ollietreend: Please import this from `@/views/Dashboard` rather than `./views/Dashboard.vue`. This is consistent with the way other modules are imported within the codebase.

Things to note here:

- `@` is an alias for the root directory of our project, so it's not dependent on the current directory of the file (which `.` is)
- we don't need to include the `.vue` file extension

hosting/apply-admin/src/router.js
```diff
@@ -14,12 +14,12 @@ const router = new Router({
...
    {
       path: '*',
-      redirect: '/',
+      redirect: '/dashboard',
     },
     {
-      path: '/',
-      name: 'home',
-      component: Home,
+      path: '/dashboard',
```
ollietreend: It probably makes sense for the dashboard view to be available at `/` rather than `/dashboard`, because this is the first page users will see once they've logged in.

hosting/apply-admin/src/router.js
```diff
@@ -46,6 +50,8 @@ router.beforeEach((to, from, next) => {
   const requiresAuth = to.matched.some(x => x.meta.requiresAuth);
   const isSignedIn = store.getters.isSignedIn;
 
+  document.title = `${to.meta.title} | Judicial Appointments Commission`;
```
ollietreend: I think the `document.title` change might be better suited in a [global `afterEach` hook](https://router.vuejs.org/guide/advanced/navigation-guards.html#global-after-hooks) instead of the `beforeEach` where we check for auth:

- It will never affect navigation. Where `beforeEach` can still cancel the navigation or cause a redirect, `afterEach` doesn't have a `next()` callback so it can't cancel or redirect – the navigation has already happened by this point.
- We should probably only change the document title when we're sure we'll be displaying that route. Currently it sits above the auth check, so document title could be set, and then immediately redirect to sign in and change the title again.
- In terms of code readability, it would be better that this happens in its own function with its own clear responsibility. Changing the document title doesn't feel like it belongs in the same place as an auth check – so right now I think we're mixing multiple responsibilities into the same function.

Ideally the `afterEach` hook would also have a small comment (similar to the one above `beforeEach`) just to explain that this is where the document title is set according to the current route.

hosting/apply-admin/tests/unit/journeys/signin.spec.js
```diff
@@ -49,24 +44,31 @@ describe('Sign in journey', () => {
     });
 
     describe('when going to page that does not exist', () => {
-      it('redirects to home page', () => {
+      it('redirects to the dashboard page and passes the title in meta', () => {
         router.push('/whatever');
-        expect(subject.vm.$route.path).toBe('/');
+        expect(subject.vm.$route.path).toBe('/dashboard');
+        expect(subject.vm.$route.meta.title).toBe('Dashboard');
```
ollietreend: Are you trying to test if the `document.title` is set correctly, or if the route meta contains the correct title? This test alone doesn't assert that the document title is updated – only that the route meta data is correct.

I can see the test below which check that `document.title` contains 'Judicial Appointments Commission'. I expected this test to be doing something similar – which will test that the document title is set, and therefore the route meta is correct.

hosting/apply-admin/src/views/Dashboard.vue
```diff
@@ -0,0 +1,33 @@
...
+      v-for="vacancy in vacancies"
+      :key="vacancy.jac_ref"
+      class="govuk-list"
+    >
+      <router-link
+        tag="li"
+        to="/dashboard"
+        class="govuk-link"
+      >
+        <a>{{ vacancy.jac_ref }} {{ vacancy.title }}</a>
```
ollietreend: [`<router-link>`](https://router.vuejs.org/api/#router-link) renders as an `<a>` element, so no need to wrap the link text in an `<a>`. Right now I think this will produce nested links.

hosting/apply-admin/src/main.js
```diff
@@ -3,7 +3,6 @@ import App from './App.vue';
 import router from './router';
```
ollietreend: I know this is unrelated to this PR, but I've just noticed that `router` and `store` are imported with `./` rather than `@/`.

Would you mind changing this as part of this PR, so that we're cleaning up tech debt as we find it?

hosting/apply-admin/src/main.js
```diff
@@ -1,9 +1,8 @@
 import Vue from 'vue';
-import App from './App.vue';
-import router from './router';
-import store from './store';
+import App from '@/App.vue';
```
ollietreend: Please also remove the `.vue`

hosting/apply-admin/tests/unit/journeys/signin.spec.js
```diff
@@ -3,70 +3,103 @@ import App from '@/App';
...
-      email: 'user@judicialappointments.digital',
-    };
 
     beforeEach(() => {
       store.dispatch('setCurrentUser', user);
     });
 
     describe('when going to page that does not exist', () => {
-      it('redirects to home page', () => {
+      it('redirects to the dashboard page and passes the title in meta', () => {
```
ollietreend: This description is now inaccurate – "and passes the title in meta". Let's just remove that from the description?

hosting/apply-admin/src/store/vacancies.js
```diff
@@ -5,7 +5,7 @@ export default {
   namespaced: true,
   actions: {
```
tatyree: @alexbokii here's what I'd try:

```js
    bind: firestoreAction(async context => {
      await auth().onAuthStateChanged(async user => {
        return await context.bindFirestoreRef('vacancies', db.collection('vacancies').orderBy('openAt'));
      });
    })
```

hosting/apply-admin/tests/unit/journeys/signin.spec.js
```diff
@@ -3,70 +3,103 @@ import App from '@/App';
...
+      it('redirects to the dashboard page', () => {
         router.push('/sign-in');
-        expect(subject.vm.$route.path).toBe('/');
+        expect(subject.vm.$route.path).toBe('/dashboard');
       });
     });
   });
 });
+
+describe('Page titles', () => {
```
ollietreend: I'm not sure tests for the page title functionality belongs in `journeys/signin.spec.js` because I think it's directly related to testing the sign in paths. It's a new top-level `describe` block so it should probably be in its own file.

But.. for now at least it's being tested. I think this may be something to refactor later when we add more routes and tests for page titles.

So no need to change it right now.

PR #214, opened 2019-08-08
New apply app

tatyree:
Closes https://trello.com/c/uv47aVGr/278-create-a-new-vue-app-in-hosting-apply

This is not-quite barebones vue app to replace the earlier version of the application form app and provide us something to build on.  

It has vue, router, vuex, and jest installed as well as the gov-uk-frontend toolkit.  I have also ported over the GTM config and the page title.  

It does not do **anything** yet, besides show the default vue page.  I changed the default test to a simpler dummy test just to remove an fragility around it as a starting point (didn't want circle choking for no reason).


hosting/apply/README.md
```diff
@@ -0,0 +1,46 @@
...
+```
+
+### Customize configuration
+See [Configuration Reference](https://cli.vuejs.org/config/).
+
+## Browser support
+
+Works in modern browsers (Chrome, Firefox, Safari, Edge).
+
+Functional in Internet Explorer 11, although some of the Bootstrap styling looks odd. Does not work in older versions
```
ollietreend: Let's drop this paragraph about bootstrap styles from the README.

hosting/apply/package.json
```diff
@@ -0,0 +1,37 @@
...
+  "version": "0.1.0",
+  "private": true,
+  "scripts": {
+    "serve": "vue-cli-service serve",
+    "build": "vue-cli-service build",
+    "lint": "vue-cli-service lint",
+    "lint-ci": "vue-cli-service lint --no-fix",
+    "test": "TZ='Europe/London' vue-cli-service test:unit",
+    "test-ci": "TZ='Europe/London' vue-cli-service test:unit --ci --runInBand",
+    "test:unit": "vue-cli-service test:unit"
```
ollietreend: Could we remove the `test:unit` command since we also have `test`?

They both essentially do the same thing (minus setting the timezone), and we've standardised on simply `test` in our other apps. It's one of the Vue defaults we've consciously diverged from.

hosting/apply/tests/unit/dummy.spec.js
```diff
@@ -0,0 +1,5 @@
+describe('Dummy test', () => {
+  it('makes circleci happy ', () => {
+    expect(true).toBeTruthy;
```
ollietreend: I don't _think_ (?) this assertion actually executes, because it should be `toBeTruthy()` not `toBeTruthy`...

But CircleCI is happy, so no need to change it. It's a no-op test and it passes as is.

hosting/apply/.eslintrc.js
```diff
@@ -0,0 +1,17 @@
+module.exports = {
```
ollietreend: Do we want to go ahead with the `eslint` config from our other apps and standardise on it now?

Or should we save that for another ticket?

Happy either way.

hosting/apply/src/App.vue
```diff
@@ -1,76 +1,29 @@
...
+#nav {
+  padding: 30px;
+  a {
+    font-weight: bold;
+    color: #2c3e50;
+    &.router-link-exact-active {
+      color: #42b983;
+    }
+  }
+}
```
alexbokii: I think we can remove these styles and leave <style> empty, as we want to keep all custom styles in main.scss file.
But we can leave it for now and remove this later when we'll be working on home page component.

hosting/apply/.eslintrc.js
tatyree: Let's leave it for now as I'd like to get it moving.

PR #215, opened 2019-08-09
Move init to onAuthStateChanged and check for isSignedIn value

alexbokii:
As we don't want to bind vacancies if a user is not logged in, init is moved to onAuthStateChanged method. Also, added a check for isSignedIn status and tests.

**Change log:**
1) Move "init" to main.js into an onAuthStateChanged method
2) In init action before dispatching 'vacancies/bind' we check for user status

### DIFF
--- a/hosting/apply-admin/tests/unit/store/store.spec.js
+++ b/hosting/apply-admin/tests/unit/store/store.spec.js
@@ -27,7 +27,7 @@ describe('Vuex store', () => {
       expect(dispatch).toHaveBeenCalled();
     });
 
-     it('not calling dispatch if user is not signed in', () => {
+     it('does not call dispatch if user is not signed in', () => {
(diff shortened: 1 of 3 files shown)


hosting/apply-admin/tests/unit/store/store.spec.js
```diff
@@ -0,0 +1,38 @@
...
+  describe('init action', () => {
+    it('calls dispatch if user is signed in', () => {
+      store.getters.isSignedIn = true;
+      const dispatch = jest.fn().mockResolvedValue();
+      store.actions.init({ dispatch });
+
+      expect(dispatch).toHaveBeenCalled();
+    });
+
+     it('not calling dispatch if user is not signed in', () => {
```
tatyree: it('does not call dispatch...

PR #227, opened 2019-11-18
send exercise started email to logged-in admin user

2-u:
when an Exercise is created.

Also created a test which creates a dummy Exercise in firestore and notifies my email address:
martin.raymundo@judicialappointments.digital

You can trigger it from your browser:
https://us-central1-digital-platform-staging.cloudfunctions.net/testExerciseNew

After you trigger it, you can check Firestore 'exercises' document here (sort by: createdAt, desc):
https://console.firebase.google.com/project/digital-platform-staging/database/firestore/data~2Fexercises

You can also check the logs here:
https://console.firebase.google.com/project/digital-platform-staging/functions/logs?severity=DEBUG


functions/exercises/testExerciseNew.js
```diff
@@ -0,0 +1,24 @@
+const functions = require('firebase-functions');
```
warrensearle: We should have a task to remove these functions for testing / internal use.
Or perhaps keep them but exclude them from being deployed to production.

functions/sharedServices.js
```diff
@@ -0,0 +1,36 @@
+const functions = require('firebase-functions');
```
warrensearle: Like the approach to have shared services

functions/exercises/testExerciseNew.js
2-u: You're right. I'll remove this test now.

PR #231, opened 2019-11-21
Add created at attribute on exercise create

2-u:
This includes refactoring of index.js, as we want to encourage making new cloud functions in separate files.

The main change here is that before, we were adding the 'createdAt' field to the Exercise document via supplying the 'createdAt' field in the 'apply-admin' Vue app. Now we add 'createdAt' field when the Exercise document onCreate method is triggered.


warrensearle:
Need to fix lint errors by running `npm run lint` locally

PR #239, opened 2019-11-27
Introducing slog slack logger

2-u:
Since show and tell yesterday, people have been entering data into
the live Production database. It is not completely ready yet, but we
need a little monitoring in place, as we will be changing the document
structure drastically as we build out the new JAC system replacing JARS.

Instead of using console.log, use slog, which does 2 things:
- do a simple console.log
- writes the same string in console.log into a private slack channel #prod-alerts

### DIFF
--- a/functions/sharedServices.js
+++ b/functions/sharedServices.js
@@ -67,14 +67,14 @@ const slog = async (msgString) => {
   // firebase functions:config:set slack.url="YOUR_SLACK_INCOMING_WEBHOOK_URL"  
   const slackUrl = functions.config().slack.url;
 
-  data = {
+  const data = {
     text: msgString,
-  }
+  };
   
   // we wait for the axios.post Promise to be resolved
   const result = axios.post(slackUrl, data);
   return result.data;
-}
+};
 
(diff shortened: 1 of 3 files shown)


functions/sharedServices.js
```diff
@@ -53,9 +54,32 @@ const getData = async (collectionName, docId) => {
...
+  // Set slack.url in firebase functions like this:
+  // firebase functions:config:set slack.url="YOUR_SLACK_INCOMING_WEBHOOK_URL"  
+  const slackUrl = functions.config().slack.url;
+
+  const data = {
+    text: msgString,
+  };
+  
+  // we wait for the axios.post Promise to be resolved
+  const result = axios.post(slackUrl, data);
```
warrensearle: does this need an `await`?

2-u: It can have await, but I was thinking if posting to the slack url might take long, this would block. So if we take out 'await', we stay async.

2-u: On second thought, I'll put `await` in. It makes better sense.

Issue #314, opened 2020-07-17
Support manual triggers for virus scanning


crablab:
We want to be able to run our virus scans across all the files we currently have (that haven't been scanned). To do this we need a function that will let us trigger a scan across an entire bucket. 

The function should take a parameter to either ignore or rescan files that already have scan metadata (#313).

crablab:
This is currently in progress using Cloud Tasks to trigger the GAE container.

PR #329, opened 2020-07-27
#314 Enable deployment of malware functions

warrensearle:
A few changes to enable deployment of malware scanner functions:
- Update package scripts to deploy to develop
- Update deploy script to only deploy malware functions
- Update functions index to only define malware functions

Note have excluded other functions from being deployed. This is fine for now. There's a refactor going on and when that's checked in we'll rebase and tidy.

### DIFF
--- a/functions/index.js
+++ b/functions/index.js
@@ -3,70 +3,13 @@
  *  create your functions in separate files
  */
 
-exports.backupFirestore = require('./backup/firestore');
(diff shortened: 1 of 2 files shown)


crablab:
Hey @warrensearle  

Thanks for looking into this! 

Will this pick up the trigger required for the function to run? When deploying manually I've had to specify this, but I'm not sure if that's required here?

PR #423, opened 2020-10-07
add inviteOnly property and update rules to show invitations to only …

tomlovesgithub:
…invited candidates


warrensearle (approved):
Looks good to me 👍 
Lets go through rules unit tests in the near future

PR #444, opened 2020-10-30
69/add hard limit to assessment

tomlovesgithub:


functions/shared/factories.js
```diff
@@ -114,6 +114,9 @@ module.exports = (CONSTANTS) => {
       fileRef: '',
       status: 'draft',
     }
+    if (exercise.independentAssessmentsHardLimitDate) {
+      assessment.hardLimit = exercise.independentAssessmentsHardLimitDate
```
warrensearle, line 118: Let's use `hardLimitDate` rather than `hardLimit`

warrensearle (changes requested):

warrensearle (approved):
👍

Issue #72, opened 2019-03-14
Implement an email action handler in the Vue app


ollietreend:
So the user's return journey to the website is more seamless.

This will be used when users click an action link in an email – e.g. reset password.

PR #445, opened 2020-11-02
#72 bad link fix - when there's an error on the link the page hangs

lloback:
THe function was not created thinking of a bad link. This fix it.


warrensearle (approved):
Good 👍

PR #454, opened 2020-11-17
Change rules for LOGS and changing the assessor

lloback:
This PR changes permissions for two tasks that I was working on:

LOGS
Change ASSESSORS

### DIFF
--- a/database/firestore.rules
+++ b/database/firestore.rules
@@ -121,15 +121,15 @@ service cloud.firestore {
       allow write: if userIsAuthenticated() && userIsJAC();
     }
 
-    match /logs/{id}/{document=**} {
-      // allow candidates update their own logs
-      allow get, read, create, write, update: if userIsAuthenticated();
+    match /logs/{document=**} {
+      allow read, write: if userIsAuthenticated();
+      allow read: if userIsAuthenticated() && userIsJAC();
     }
 
     match /assessments/{assessmentId} {
       // allow admins full read and update permission
       allow read, get: if userIsAuthenticated() && userIsJAC();
-      allow update, create: if userIsAuthenticated() && userIsJAC();
+      allow update, create: if userIsAuthenticated();
 
       // allow assessor to read and update own documents
       allow read: if userIsAuthenticated() && userHasEmail(resource.data.assessor.email);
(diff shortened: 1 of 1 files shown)


database/firestore.rules
```diff
@@ -121,10 +121,15 @@ service cloud.firestore {
       allow write: if userIsAuthenticated() && userIsJAC();
     }
 
+    match /logs/{id}/{document=**} {
+      // allow candidates update their own logs
+      allow get, read, create, write, update: if userIsAuthenticated();
```
warrensearle, line 126: There are un-necessary verbs here:
`read` and `write` are high-level and cover the relevant lower level verbs
`read` implies `list` and `get`
`write` implies `create`, `update`, `delete`

database/firestore.rules
```diff
@@ -121,10 +121,15 @@ service cloud.firestore {
       allow write: if userIsAuthenticated() && userIsJAC();
     }
 
+    match /logs/{id}/{document=**} {
+      // allow candidates update their own logs
```
warrensearle, line 125: This doesn't restrict candidate to only updating their own log

warrensearle (changes requested):

lloback:
The only thing that seems to works here is:

```
match /logs/{document=**} {
      // allow candidates update their own logs
      allow read, write: if userIsAuthenticated();
      allow read: if userIsAuthenticated() && userIsJAC();
    }
```

lloback:
I've tried different combinatiosn and it doesn't work:

```
match /logs/login/{id} {
      allow read, write: if userIsAuthenticated() && currentUser() == id; // DON"T WORK
      allow read: if userIsAuthenticated() && userIsJAC();
    }
```

```
match /logs/{document=**}/{id} {
      allow read, write: if userIsAuthenticated() && currentUser() == id; // DON"T WORK
      allow read: if userIsAuthenticated() && userIsJAC();
    }
```

warrensearle:
I think the following should do the trick:
```
match /logs/login/{id}/{logId} {
      // allow candidates update their own logs
      allow read, write: if userIsAuthenticated() && currentUser() == id;
      allow read: if userIsAuthenticated() && userIsJAC();
    }
```

database/firestore.rules
lloback: this is implemented.

warrensearle (approved):
👍

PR #464, opened 2020-12-01
#457 Security rules to allow application extension

warrensearle:
## Tests ##
```
Close date | Extension date | Able to save? | Result
Future     | None           | Yes           | PASS
Past       | None           | No            | PASS
Past       | Future         | Yes           | PASS
Past       | Past           | No            | PASS
Future     | Future         | Yes           | PASS
Future     | Past           | Yes           | FAIL
```

**Note**
Currently if we have an extension date it will always override the exercise close date. This means that if an exercise is open but an application has an extension date set in the past then the candidate will be unable to amend their application. This might actually be useful behaviour however I have marked it as a FAIL in the tests above.


lloback (approved):
This is JUST WHAT I NEEDED! Thanks. Cheers.

PR #477, opened 2020-12-18
admin#1110 Update diversity report to include withdrawn

warrensearle:
Also update outreach report to include withdrawn
Fix lint errors


lloback (approved):
ok

PR #547, opened 2021-05-14
#546 Bugfix: Should be able to initialise a QT

### LINKED ISSUE #546: Should be able to initialise a Qualifying Test
There appears to be a problem initialising a Qualifying Test. 

From admin the console shows that the `initialiseQualifyingTest` cloud function is returning `false` where we would expect it to initialise the QTs and return the number that were initialised.

warrensearle:
Closes #546 

Callable function `initialiseQualifyingTest` was returning the following error:

```
Cannot use "undefined" as a Firestore value (found in field "qualifyingTest.isTieBreaker"
```

This PR fixes that.


mbrookeswebdev (approved):

HalcyonJAC (approved):
Looks good to me :)

PR #557, opened 2021-05-20
Update firestore indexes

warrensearle:
Updates our firestore index definitions file so we can deploy it to other environments.
Plus adds an npm run script so we can update them again in the future: `npm run indexes:export`


mbrookeswebdev (approved):

HalcyonJAC (approved):
love it! :)

PR #565, opened 2021-05-21
check assessor emails

HalcyonJAC:
Updated `sendAssessmentRequests` and `sendAssessmentReminders` callable functions to check assessor email addresses before sending emails to them and return errors if any are invalid (i.e. instead of sending the emails).


lloback (approved):

warrensearle (approved):
Awesome 👍

PR #570, opened 2021-05-28
Feature/admin 1312 notifications

mbrookeswebdev:
Ready to code review (please do it together with [#1327](https://github.com/jac-uk/admin/pull/1357), [#773](https://github.com/jac-uk/apply/pull/791) and [#627).](https://github.com/jac-uk/digital-platform/pull/627)


lloback:
@mbrookeswebdev let us know when this is ready for review.

mbrookeswebdev:
This work needs to be integrated with Staged Application feature after I've finished preparation for access to JO API.

mbrookeswebdev:
Ready to code review (please do it together with #[1327](https://github.com/jac-uk/admin/pull/1357) and #[773](https://github.com/jac-uk/apply/pull/791)).

functions/callableFunctions/updateCharacterChecksStatus.js
```diff
@@ -0,0 +1,18 @@
...
+    applicationRecordId: { required: false },
+    exerciseId: { required: true },
+  }, data)) {
+    throw new functions.https.HttpsError('invalid-argument', 'Please provide valid arguments');
+  }
+  if (!context.auth) {
+    throw new functions.https.HttpsError('failed-precondition', 'The function must be called while authenticated.');
+  }
+  return await updateCharacterChecksStatus(data);
+});
```
warrensearle, line 18: Looking at related work here https://github.com/jac-uk/apply/pull/791 I think this might be better as a background function rather than callable function. DM me and we can discuss further.

warrensearle (changes requested):

functions/callableFunctions/updateCharacterChecksStatus.js
mbrookeswebdev: @warrensearle Thanks Warren, I'll have a look at #791 and make changes as requested.

mbrookeswebdev: Done

functions/actions/applications/onUpdate.js
```diff
@@ -45,7 +47,24 @@ module.exports = (config, firebase, db) => {
...
+            'characterChecks.status': 'completed',
+            'characterChecks.completedAt': firebase.firestore.Timestamp.fromDate(new Date()),
+          });
+          return true;
+        } catch (e) {
+          console.error(`Error updating application record ${applicationId}`, e);
+          return false;
+        }
+      }
+    }
```
warrensearle, line 67: We should check existence of `.characterChecks` before looking at `.characterChecks.status` as not all applications will have the `characterChecks` field

warrensearle (changes requested):
Looks good, just one small tweak 😄

functions/actions/applications/onUpdate.js
mbrookeswebdev: I've added checks

warrensearle (approved):
Awesome 👍

functions/shared/factories.js
```diff
@@ -1,13 +1,48 @@
...
   };
 
+  function newNotificationCharacterCheckRequest(firebase, application, type, exerciseMailbox, exerciseManagerName, dueDate) {
+    let templateId = '';
+    let templateName = '';
+    if (type === 'request') {
+      templateId = '5a4e7cbb-ab66-49a4-a8ad-7cbb399a8aa9';
+      templateName = 'Character Check Request';
+    } else {
+      templateId = '5a4e7cbb-ab66-49a4-a8ad-7cbb399a8aa9'; // need a different reminder template
```
HalcyonJAC, line 19: This template ID is the same as the above, but the code reads like it should be different. Is there a better way of doing this instead of inserting IDs into the code itself?

functions/shared/factories.js
```diff
@@ -1,13 +1,48 @@
...
+      email: application.personalDetails.email,
+      replyTo: exerciseMailbox,
+      template: {
+        name: templateName,
+        id: templateId,
+      },
+      personalisation: {
+        exerciseName: application.exerciseName,
+        dueDate: dueDate,
+        urlRequired: 'https://apply.judicialappointments.digital/sign-in',
```
HalcyonJAC, line 32: Does this need to be environment specific?

HalcyonJAC (changes requested):

functions/shared/factories.js
mbrookeswebdev: @HalcyonJAC we are in the process of setting up two different templates for a request and a reminder emails, so there will be two different IDs. Do you think we should set this up differently? I followed the IA/notifications pattern we already have following Warren's guidance.

functions/shared/factories.js
mbrookeswebdev: I've made a change.

functions/shared/factories.js
HalcyonJAC: @mbrookeswebdev I guess as long as we don't forget to update the ID prior to deployment then that's fine! 😄 

If this is the agreed practise then no problem! I think personally, I would avoid storing IDs like this within the code itself and have a single place where IDs for templates are stored. I would potentially use something like a separate file which contains the templateName and the corresponding ID. Then you could use something like `getTemplateByName('characterCheckRequest')` which returns the relevant info on the template (ID and name) and keeps all that information in a single place and is reusable across the platform. Something to consider as a refactoring exercise in the future perhaps! 

Otherwise I tested it and it worked great 💥

HalcyonJAC (approved):

tomlovesgithub (approved):

PR #575, opened 2021-06-02
#574 Enable application form to be configured

warrensearle:
Copies the new `applicationContent` field from `exercises` to `vacancies`. This helps to enable Apply form to be configurable.

Plus adds an package script for `npm run emulators:firestore`


HalcyonJAC (approved):
OK - This is on the wrong pull request - I'm just looking into it now...

PR #576, opened 2021-06-02
Retrieve ASC titles to vacancy not application

tomlovesgithub:
As part of 1271 i've removed the ASC titles from the application objects so they can be updated without having to update any preexisting applications. 
Here i ensure the same data is carried through in 'application converter'


warrensearle (approved):
Code change looks fine to me.

Issue #592, opened 2021-06-18
Support having an application process


warrensearle:
We need to update the digital platform to enable application content to be added by candidates after the exercise has closed and at relevant stages of the application process.

- [x] Functions. Ensure `vacancy` stays up to date with changes in exercise state
- [x] Firestore security. Update rules to allow candidates to add extra content to their application **when** the vacancy requires it

PR #593, opened 2021-06-18
#592 Support application process

warrensearle:
This PR covers the following:

Ensures `vacancy` documents to stay up to date with all exercise changes. Previously the vacancy was not updated after the exercise had been approved, as no changes were happening. Now the exercise changes state and this should be synced to the vacancy.

Updates our security rules for firestore to allow candidates to add extra content to their application **when** the vacancy requires it. Have included main tests covering this scenario.

Fixes an incompatibility with the new `@firebase/rules-unit-testing` package where it allows the creation of timestamps that are incompatible with firestore. So have updated our helper to use `firebase-admin` to create timestamps instead.

### DIFF
--- a/database/firestore.rules
+++ b/database/firestore.rules
@@ -101,6 +101,12 @@ service cloud.firestore {
       return applicationOpenDate <= request.time
 				&& request.resource.data.get("dateExtension", applicationCloseDate) > request.time;
     }
+    function exerciseNeedsMoreInformation(exerciseId) {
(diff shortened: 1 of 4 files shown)


tomlovesgithub (approved):
Cool! Bonus points for the tests!

Is there planned UI to inform candidates that details of an exercise have changed?
Do you think the existing data is sufficient to do this?

HalcyonJAC (approved):

mbrookeswebdev (approved):
Great work Warren 👍

PR #597, opened 2021-06-26
admin #1375 - Updates to Storage and Firestore rules

tomlovesgithub:
Rules from current develop env on firebase.


warrensearle (approved):
This includes quite a few changes but all looks ok.

PR #602, opened 2021-06-29
Rules changes for Microsoft auth

tomlovesgithub:


warrensearle (approved):
Hey @tomlovesgithub approved as this reflects what has been deployed already (!) however further refinement is needed, as discussed.

Issue #483, opened 2021-01-04
Cloud function to update status based on cut-off score


warrensearle:

KateMJAC:
Discussed as part of the QT II epic today. It was agreed that this ticket was a priority to be included in April's work.

JeremykJAC:
More detail needed :)

PR #603, opened 2021-06-30
#483 Cloud function to update status based on cut-off score

HalcyonJAC:
## What's included?

The ability to capture a cut-off score from within the QT Score Report and set the status of applicationRecords that are lower / higher than X.

The ability to change individual (or multiple) application statuses by selecting the application and clicking Set status

## Who should test?
✅ Product owner
✅ Developers
✅ UTG

### Cut off scores
- Go to QT Report and click 'Set pass mark'
- Enter the cut off score and set the statuses
- Check the statuses of the applicationRecords (note: I can't find the application status anywhere in the UI which may need to be fixed as part of a different task)

### Update status
- Click on an individual score from the QT report
- Select one or more applications
- Click 'Set status'
- Select a status and continue
- Check the status(es) of the applicationRecords relating to the application you updated. (note: I can't find the application status anywhere in the UI which may need to be fixed as part of a different task)

https://user-images.githubusercontent.com/79906532/124025875-e60a3200-d9e8-11eb-8a80-31b47ee1bda9.mp4

## Additional context

---
PREVIEW:DEVELOP
_can be OFF, DEVELOP or STAGING_


functions/actions/qualifyingTests/cutOffScoreQualifyingTest.js
```diff
@@ -0,0 +1,42 @@
...
+      //set the status
+      let status = null;
+      if (application.score >= params.cutOffScore) {
+        status = params.cutOffPassStatus;
+      } else {
+        status = params.cutOffFailStatus;
+      }
+
+      // get existing applicationRecord and add update command to batch
+      await updateApplicationRecords(application.applicationId, status);
```
warrensearle, line 20: Hey @HalcyonJAC rather than updating the global  commands array within the `updateApplicationRecords` function / sub routine I wonder if the function should return a command (if relevant) which is then added to a more local commands array.
Is there a risk we will be re-running commands from a previous call to `updateStatuses`?
Other than that I think it’s perfect :)

warrensearle (changes requested):
See comment... I think this needs tweaking

functions/actions/qualifyingTests/cutOffScoreQualifyingTest.js
HalcyonJAC: Gotcha - great point. I forget that global variables can be available across multiple requests!

tomlovesgithub (approved):

warrensearle (approved):
👍

Issue #419, opened 2020-10-06
Populate application records with diversity data


warrensearle:
We need to ensure diversity stats are copied across to the `applicationRecords` documents for each existing exercise.

There are two requirements here:

1. Ensure new `applicationRecords` include the diversity stats
2. Ensure existing `applicationRecords` are updated to include diversity stats

warrensearle:
This is blocking admin#930 from being tested

warrensearle:
The groundwork has been done for this. We have a nodeScript called `addDiversityToApplicationRecords` which can be manually run to do that!

It will be needed for QTs Results epic however have removed from the Statistics epic (admin#849) as it is no longer a requirement there.

When working on QT Results we will implement best way to ensure application records have up to date diversity data

JeremykJAC:
Speak with Warren

warrensearle:
See PR #604 for additional discussion

PR #604, opened 2021-06-30
#419 Populate application records with diversity data

HalcyonJAC:
## What's included?

When a new applicationRecord document is created in firestore, it now includes the diversity data

## Who should test?
✅ Developers

### Cut off scores
- Apply for a vacancy then process the applications from the admin.
- Check that the diversity data is included within the applicationRecord document

## Additional context

---
PREVIEW:DEVELOP
_can be OFF, DEVELOP or STAGING_

### DIFF
--- a/functions/shared/factories.js
+++ b/functions/shared/factories.js
@@ -148,6 +148,73 @@ module.exports = (CONSTANTS) => {
     return assessment;
   }
 
+  function newDiversityFlags(application) {
+    const applicationData = application.equalityAndDiversitySurvey ? application.equalityAndDiversitySurvey : application;
+    const data = {
+      gender: applicationData.gender,
+      ethnicity: null,
+      disability: applicationData.disability,
+      professionalBackground: {
+        barrister: null,
+        cilex: null,
+        solicitor: null,
+        other: null,
+        preferNotToSay: null,
+      },
+      socialMobility: {
+        attendedUKStateSchool: null,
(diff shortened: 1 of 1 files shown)


warrensearle (approved):
This PR looks good to me. 👍 
As part of the parent ticket we will also need to address 
 - keeping the data up to date
 - populating existing `applicationrecords` (back-filling)

Please clarify whether this will be included here or in follow-up PR(s)

HalcyonJAC:
@warrensearle Does an application's diversity data ever change once the applications have been processed? I had assumed it doesn't but if it does I can add a db trigger to update the applicationRecord. 

Also updating the existing records - we could run the existing nodeScript for this, but what is the process for doing this / other one-off migrations / data changes on the production environment? We could have a button per exercise that generates the diversity data, but it is only applicable for old exercises so could be over-kill

tomlovesgithub (approved):

mbrookeswebdev (approved):

warrensearle:
Hey @HalcyonJAC we should assume that the data might be updated even after the team have started processing applications, so yes please add to the `onApplicationUpdate` background function to check for the relevant changes and update `applicationRecord` (if it exists).
This would be a good event to log so we can track how often this happens.

warrensearle:
With regards back-filling data we should only need to run the existing node script once then the new functions will take over. DM me and we can go through existing process for this.

PR #605, opened 2021-07-01
Update package script to deploy all functions

warrensearle:
Our clouds functions are now all in source control therefore we no longer need to specify the individual functions we wish to deploy. We can simply deploy whatever is listed in our functions `index` file.
(I am excited at least as it has been a pain to manage which functions we're including/excluding in our deploy)


HalcyonJAC (approved):

mbrookeswebdev (approved):
👍

Issue #473, opened 2020-12-14
Purge backups after 30 days


warrensearle:
Our backups for firestore and authentication data are saved in firebase storage. At present we are not removing old backups. Our storage requirements are therefore growing daily.

This task is to create a scheduled function to delete backups when they are 30 or more days old.

KateMJAC:
Assigned to Drie in March WO.

KateMJAC:
Hello @warrensearle is it possible to get an update on this ticket?

PR #626, opened 2021-07-23
#473 purge backups after 30 days

HalcyonJAC:
Code review only please. The idea is that after 30 days, old backups are deleted from storage.

There are 2 different methods for this.

Authentication exports to a single file so processing the deletes is quick and easy. We check for anything older than 30 days.

Firestore exports to a folder with lots of files inside it each time. Here we are simply deleting anything that starts with the date today - 30 days, so each day, the export from 30 days ago is deleted.


mbrookeswebdev (approved):

warrensearle (approved):
Looks good to me 👍

KateMJAC:
Both suppler and in-house devs have approved. Happy to merge.

PR #632, opened 2021-08-12
#1380 - User Roles Phase 1

HalcyonJAC:
Please see https://github.com/jac-uk/admin/pull/1470 for further info


functions/actions/userRoles.js
```diff
@@ -0,0 +1,267 @@
+const userRoles = require('firebase-admin');
```
warrensearle, line 1: Could this be called `admin` instead of `userRoles`?
Also I think rather than requiring it here we could pass it into the module, similar to the way `db` is injected.

functions/actions/userRoles.js
```diff
@@ -0,0 +1,267 @@
...
+const { getDocument, getDocuments } = require('../shared/helpers');
+const PERMISSIONS = require('../shared/permissions');
+
+module.exports = (db) => {
```
warrensearle, line 5: `admin` could be passed in here

functions/actions/userRoles.js
```diff
@@ -0,0 +1,267 @@
...
+    try {
+      // get all users
+      const users = await userRoles.auth().listUsers();
+
+      for(const user of users.users) {
+        let isJacAdmin = false;
+        let isJACEmployee = false;
+        for (const provider of user.providerData) { // users can authenticate on both admin and apply with same email
+          if(provider.providerId === 'google.com' || provider.providerId === 'microsoft.com') {
+            isJacAdmin = true; // user has authenticated successfully with google or microsoft
```
warrensearle, line 38: We should restrict this to only `@judicialappointments.(digital|gov.uk)` email addresses, just to be sure.
I think currently anyone authenticated with google or microsoft would be allowed through...

warrensearle (changes requested):
Some questions / possible changes in the comments.

functions/actions/userRoles.js
warrensearle: As discussed this will mean these users are included in the list of users the team can configure roles for.
Let's leave it as you have it in case the team want to see which users outside JAC have tried logging in to admin.
Plus in case the team want to assign a role to non-JAC email addresses.

mbrookeswebdev (approved):
👍

warrensearle:
As discussed @HalcyonJAC is going to amend so we are passing firebase (or auth) from `helpers` into the `userRoles` module

warrensearle:
Just the following to do then we're done
> As discussed @HalcyonJAC is going to amend so we are passing firebase (or auth) from `helpers` into the `userRoles` module

warrensearle:
@HalcyonJAC @JeremykJAC any update on this one?

warrensearle:
@HalcyonJAC I've made the code changes so will merge this work

PR #646, opened 2021-09-29
#1495 Re-organised data as per application layout, changed some incorrect info and added missing fields

mbrookeswebdev:
To test: 
Go to any existing exercise with applications -> click on 'Refresh'. 
All candidates with character issues should display on the page.


warrensearle (approved):
Code change looks good to me 👍 
(not tested)

PR #654, opened 2021-10-27
Feature/653 jo handover download

mbrookeswebdev:
To test: in Admin go to an exercise -> Reports -> Handover, click 'Refresh' and 'Export data' (at least one candidate has to be in Handover stage).


warrensearle (approved):
👍

PR #677, opened 2022-01-21
257 - database security rules

HalcyonJAC:
This commit is to redo a commit that was previously rolled-back


warrensearle (approved):
This looks good but lets hold off from merging until you are back from your holiday. Just in case! 😄

tomlovesgithub (approved):

PR #701, opened 2022-04-13
Fix: Event Logger

warrensearle:
Fixes a bug where cloud functions `onApplicationRecordUpdate` and `initialiseMissingApplicationRecords` were not behaving as expected.

The event logger module `logEvent` was loading the `auth` service directly rather than it being injected. This meant our app instance was effectively being initialised twice!

Have updated all instances of `logEvent` so that `auth` is now passed in.

### DIFF
--- a/nodeScripts/initialiseApplicationRecords.js
+++ b/nodeScripts/initialiseApplicationRecords.js
@@ -2,7 +2,7 @@
 
 const config = require('./shared/config');
 const { auth, firebase, app, db } = require('./shared/admin.js');
-const { initialiseApplicationRecords } = require('../functions/actions/applicationRecords')(config, auth, firebase, db);
+const { initialiseApplicationRecords } = require('../functions/actions/applicationRecords')(config, firebase, db, auth);
 
 const main = async () => {
   return initialiseApplicationRecords({
--- a/nodeScripts/initialiseApplicationRecords.js
+++ b/nodeScripts/initialiseApplicationRecords.js
@@ -1,12 +1,12 @@
 'use strict';
 
 const config = require('./shared/config');
-const { firebase, app, db } = require('./shared/admin.js');
-const { initialiseApplicationRecords } = require('../functions/actions/applicationRecords')(config, firebase, db);
+const { auth, firebase, app, db } = require('./shared/admin.js');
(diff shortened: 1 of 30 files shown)


nodeScripts/initialiseApplicationRecords.js
```diff
@@ -1,12 +1,12 @@
...
 const config = require('./shared/config');
-const { firebase, app, db } = require('./shared/admin.js');
-const { initialiseApplicationRecords } = require('../functions/actions/applicationRecords')(config, firebase, db);
+const { auth, firebase, app, db } = require('./shared/admin.js');
+const { initialiseMissingApplicationRecords } = require('../functions/actions/applicationRecords')(config, auth, firebase, db);
 
 const main = async () => {
-  return initialiseApplicationRecords({
-    exerciseId: '8CIlAsDbtMfr2vnfjmYh',
+  return initialiseMissingApplicationRecords({
```
tomlovesgithub, line 8: is it worth making another nodeScript for this? or maybe changing this one back to match the filename? (just seems potentially missable in the future)

tomlovesgithub (approved):

nodeScripts/initialiseApplicationRecords.js
warrensearle: Yes you're right @tomlovesgithub, I shall change the function back to match the script name. That said before running a node script you would usually check the content contains the params you want to be running. Plus the node scripts aren't deployed, they are just used by us devs to test out the code we've written... so we can be slightly less precise with them 😄  - e.g. feel free to add a script called `toms-script` and have it run whichever actions you like

HalcyonJAC (changes requested):
Hi Warren, could you also modify the line 6 and 7 in the file `./functions/actions/applications/applications.js`. It will be imported by `createTestApplications.js` so need to pass `auth` into it as well.

Before:

```js
module.exports = (config, firebase, db) => {
  const { initialiseApplicationRecords } = require('../../actions/applicationRecords')(config, firebase, db);
```

After:

```js
module.exports = (config, firebase, db, auth) => {
  const { initialiseApplicationRecords } = require('../../actions/applicationRecords')(config, auth, firebase, db);
```

HalcyonJAC (approved):

PR #702, opened 2022-04-13
Fix: Application record personal details defaults

warrensearle:
Ensures `applicationRecord` documents can be created even when the `application` is missing some `personalDetails` data.
(with the default values the write to firestore was failing)


tomlovesgithub (approved):
Nice! this platform stuff can be hard to resolve sometimes - ill bet this one is a stitch in time!

HalcyonJAC (approved):
Code looks great to me 👍

Issue #703, opened 2022-04-21
User Roles Part 2 (Implementation)


HalcyonJAC:

andrew-isaac:
@JeremykJAC  - As this is rolled out can I close the ticket?

andrew-isaac:
@JeremykJAC  - As this is rolled out can I close the ticket?

PR #705, opened 2022-04-22
#703 User Roles Part 2 (Implementation)

HalcyonJAC:
## What's included?
Implement user role and permission check on Firebase Function to secure Firestore:
1. Fixed functions `adminSetUserRole` and `adminSyncUserRolePermissions` in `functions/actions/userRoles.js`.
2. Added new function `deleteUsers` to delete users.
3. Used `auth.token.rp` to check permission.
1. Added permission check in Firestore security rules.

Please see [admin #1607 User Roles Part 2 (Implementation)](https://github.com/jac-uk/admin/pull/1610) for frontend implementation.

## How to test?
Switch to different roles on admin website ([feature/1607-user-roles-part2-implementation](https://github.com/jac-uk/admin/tree/feature/1607-user-roles-part2-implementation)) and use Postman to test the following Firebase functions:
- `adminSetUserRole` with "canChangeUserRole" permission
- `deleteUsers` with "canDeleteUsers" permission
- adminSyncUserRolePermissions

Test firestore rules: `FIRESTORE_EMULATOR_HOST=localhost:8282 npm run test:rules`
Test cloud functions: `npm run test:functions`
Permission list: https://docs.google.com/spreadsheets/d/1hbchU6SQXyEpwowBBFucNSncWhLtoPgt/edit?usp=sharing&ouid=107962156268070514376&rtpof=true&sd=true


warrensearle (changes requested):
This looks good so far however the permissions need rolling out to firestore security rules.

Right now it is using `d1`, `d2`, `d3` flags throughout however it should be using permissions corresponding to the collection - e.g. `e*` for `exercises`, `u*` for `users`

Also please remember to update the tests or even write the amended tests first 😄

warrensearle:
Hey @HalcyonJAC  just wanted to say awesome work with the security rules unit tests. 👍 

For bonus points it would be great to have the tests use the `functions/shared/permissions.js` object so that the tests at least use the developer-friendly tokens rather than the underlying short codes. 

e.g. use `rp: [ PERMISSIONS.users.permissions.canCreateRoles.value ]` instead of `rp: ['u5']`

I know it is more verbose but it falls in line with what our developers will be familiar with using in the UIs and helps ensure the permissions stay relevant to what is intended.

I suspect it may not possible but if the same could be used in the actual rules files that would be awesome.

e.g. allow `read: if hasPermission(PERMISSIONS.users.permissions.canCreateRoles.value)` instead of `allow read: if hasPermission('u5')` 

or perhaps `allow read: if hasPermission('canCreateRoles')` might be possible.

HalcyonJAC:
Hi @warrensearle, I've changed Firestore security rules and it allows developers to use this format `allow read: if hasPermission('canCreateRoles')` to assign permissions. The idea is that it will retrieve `enabledPermissions` from `roles` collection based on the `roleId` stored in `request.auth.token.r` to verify the permission.

warrensearle:
> Hi @warrensearle, I've changed Firestore security rules and it allows developers to use this format `allow read: if hasPermission('canCreateRoles')` to assign permissions. The idea is that it will retrieve `enabledPermissions` from `roles` collection based on the `roleId` stored in `request.auth.token.r` to verify the permission.

Hi @HalcyonJAC the only problem with that is it means an extra document read per request, effectively doubling our quota usage. It's not possible to import the local `permissions.js` data but it _may_ be possible to have some in-line copied version of it using local variables (`let someVar =`) in your function. However if this isn't fruitful then sticking with the inline `hasPermission('e1')` would be better than the extra document read.

HalcyonJAC:
Hi @warrensearle, I've tried to use a variable to store the `permissions` object but it seems that firestore security rules language doesn't support this syntax. So can we just use this way `hasPermission('e1')` to verify the permission?

warrensearle (approved):
Great work @HalcyonJAC thank you 👍

warrensearle (approved):
Re-approved! 💯

PR #763, opened 2022-09-12
Bugfix/Emp-flag-QT-init

tomlovesgithub:
EMP flag was assumed to be `Boolean` but after recent update can also be `'gender'` or `'ethnicity'`  
Due to this EMP candidates couldnt be added to QTs.


warrensearle (approved):
Looks good to me

HalcyonJAC (approved):

PR #776, opened 2022-10-20
#933 Application confirmation

HalcyonJAC:
## What's included?
Send the application confirmation email to applicants after applications are submitted. 

The email template:
Screenshot 2022-10-20 at 14 21 26

## Who should test?
✅ Developers

## How to test?
Run the command `npm run nodeScript sendApplicationConfirmation` in the terminal.

🟢 No risk - this is a self-contained piece of work

## Related permissions
- No permission changes required

---
PREVIEW:DEVELOP

### DIFF
--- a/functions/actions/applications/applications.js
+++ b/functions/actions/applications/applications.js
@@ -6,12 +6,13 @@ const testApplicationsFileName = 'test_applications.json';
 module.exports = (config, firebase, db, auth) => {
   const { initialiseApplicationRecords } = require('../../actions/applicationRecords')(config, firebase, db, auth);
   const { refreshApplicationCounts } = require('../../actions/exercises/refreshApplicationCounts')(firebase, db);
-  const { newNotificationCharacterCheckRequest } = require('../../shared/factories')(config);
+  const { newNotificationApplicationSubmit, newNotificationCharacterCheckRequest } = require('../../shared/factories')(config);
   const slack = require('../../shared/slack')(config);
   const { updateCandidate } = require('../candidates/search')(firebase, db);
   return {
     updateApplication,
     onApplicationCreate,
+    sendApplicationConfirmation,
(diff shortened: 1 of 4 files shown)


warrensearle (changes requested):
@HalcyonJAC this is good, just a few changes:

1. There's no need to retrieve the `application` document again as we already have it in the `dataAfter` object.
2. The `sendApplicationConfirmation` function should be amended to be solely for sending a single email as that's the functionality we are building & testing here. The current parameters indicate it could be used for sending multiple application emails but we are not building/testing for that functionality at this time.
3. Please include a flag, along the same lines as for the Character Consent email we discussed a few months ago, so that the email is only sent out once per candidate. e.g. If the application is changed from `draft` to `applied` to `draft` to `applied` then the candidate should receive one email not two. As things stand they would receive two emails.

Think that's it for now 😄

HalcyonJAC:
> @HalcyonJAC this is good, just a few changes:
> 
> 1. There's no need to retrieve the `application` document again as we already have it in the `dataAfter` object.
> 2. The `sendApplicationConfirmation` function should be amended to be solely for sending a single email as that's the functionality we are building & testing here. The current parameters indicate it could be used for sending multiple application emails but we are not building/testing for that functionality at this time.
> 3. Please include a flag, along the same lines as for the Character Consent email we discussed a few months ago, so that the email is only sent out once per candidate. e.g. If the application is changed from `draft` to `applied` to `draft` to `applied` then the candidate should receive one email not two. As things stand they would receive two emails.
> 
> Think that's it for now 😄

@warrensearle thank you for your suggestions. 😀

warrensearle (approved):
Great @HalcyonJAC thanks for making those changes ✅

tomlovesgithub (approved):

PR #783, opened 2022-11-02
Hotfix 1.29.1: Activate QT. Include dates when opening tests

warrensearle:
This fix ensures that start and end dates are copied across to each `qualifyingTestResponse` when activating a Qualifying Test.
So that any changes to dates & times are brought across to the tests.

Note: this a hotfix and we will deploy directly from this branch, as well as merge into `develop`, when it has been approved.


HalcyonJAC (approved):
👍

tomlovesgithub (approved):

Issue #786, opened 2022-11-09
Migrate qualification complete field in application collection


HalcyonJAC:
Use a new field `qualification.qualificationComplete` in the application collection instead of `qualification.qualificationNotComplete` to mark if a candidate has completed the pupillage.

warrensearle:
This is a good plan @HalcyonJAC  ✅ 
Thank you for sharing and seeking review from your peers 👍

warrensearle:
I'm ok with leaving the old field in there for existing documents. New documents will be clean; existing documents will be forwards and backwards compatible.

HalcyonJAC:
Thank you Warren, I will keep the old field `qualification.qualificationNotComplete`.

PR #790, opened 2022-11-10
#786 Migrate qualification complete field in application collection

HalcyonJAC:
## What's included?
Use a new field `qualification.completedPupillage` in the application collection instead of `qualification.qualificationNotComplete` to mark if a candidate has completed the pupillage.

### Development plan:

1. Create a node script.
2. Query for the application documents which has `qualification.qualificationNotComplete` property.
3. Iterate the results.
4. Update each document with new field `qualification.completedPupillage` based on the opposite value of `qualification.qualificationNotComplete`.
5. Go through all JAC repositories and use `qualification.completedPupillage` instead of `qualification.qualificationNotComplete`.

**6. (Will not do) Delete `qualification.qualificationNotComplete` field.** 

## Who should test?
✅ Developers

## How to test?
1. Open file `nodeScripts/migrateQualificationComplete.js`.
2. Run `npm run nodeScript migrateQualificationComplete`.
3. Modify flag `isAction` to `true` to update changes in Firestore.
4. Run `npm run nodeScript migrateQualificationComplete`.

#786 Migrate qualification complete field in application collection

### DIFF
--- a/functions/shared/converters/applicationConverter.js
+++ b/functions/shared/converters/applicationConverter.js
@@ -150,7 +150,7 @@ module.exports = () => {
           addField(data, 'Date qualified', formatDate(q.date));
         }
       }
-      if (q.qualificationNotComplete && q.details) {
+      if (!q.qualificationComplete && q.details) {
         addField(data, 'Completed pupillage', 'No');
         addField(data, 'Did not complete pupillage notes', q.details);
       }
--- a/functions/shared/converters/applicationConverter.js
+++ b/functions/shared/converters/applicationConverter.js
@@ -150,7 +150,7 @@ module.exports = () => {
           addField(data, 'Date qualified', formatDate(q.date));
         }
       }
-      if (!q.qualificationComplete && q.details) {
+      if (!q.completedPupillage && q.details) {
         addField(data, 'Completed pupillage', 'No');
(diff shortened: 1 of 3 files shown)


warrensearle (commented):
@HalcyonJAC this is looking good.

Am thinking out loud here however we could give ourselves a safety net (for the first few runs if not all or them) by creating a copy of each of the documents that will be changed by this migration: e.g. create an exact copy of the document (before changes) in a new collection called say `applications_temp`. Then make the changes and either manually check all is ok or loop through `applications_temp` and verify the latest `application` is as expected. We can delete the `applications_temp` documents and collection when we are confident all is ok.

Alternatively (or as well) we could use a cloud function e.g. `onApplicationUpdate` to validate the migration has had then desired effect (and no side effects).

tomlovesgithub (approved):

warrensearle (approved):
This looks fine to me. Sorry for the delay approving. It doesn't really need to be merged into our code base or, at least, can be removed once it has been run on all platforms.
You'll need to change the query so it runs for all applications for all exercises (or run it for each exercise one by one).
Happy to help/pair when you are actually running this if you need.

warrensearle (approved):
Good job, thanks @HalcyonJAC 👍

PR #798, opened 2022-11-25
Increase memory of virus scanning function

HalcyonJAC:
## What's included?
Increase the memory of the schedule function `scheduleScanAllFiles` to `4GB`.

Error log:
Screenshot 2022-11-25 at 12 54 26


warrensearle (approved):
Code change is fine for an immediate fix ✅ 
Can you create another task to look at making the file scanner more efficient, so it uses less memory resources

tomlovesgithub (approved):

PR #799, opened 2022-12-01
indexes from production

tomlovesgithub:
indexes from production pulled using `firebase firestore:indexes`
Contains indexes which caused QT issue on 01/12/22


HalcyonJAC (approved):

warrensearle (approved):
Good job, thanks @tomlovesgithub 
This adds one new index. The other changes have simply moved two indexes to different positions in the code

PR #818, opened 2023-01-06
Create fullNameUC for applications sorting

HalcyonJAC:
## What's included?
Modify the background function `onApplicationCreate` and `onApplicationUpdate` to update the field `_sort.fullNameUC` in the application document if the field `personalDetails.fullName` has changed.

## Who should test?
✅ Developers

## How to test?
[apply]
1. Apply for a vacancy.
2. Go to Firebase console and check if the field `_sort.fullNameUC` in the application document is populated and converted to uppercase from `personalDetails.fullName`.

[admin]
1. Find an application and change the first name or last name.
2. Go to Firebase console and check if the field `_sort.fullNameUC` in the application document is populated and converted to uppercase from `personalDetails.fullName`. 

🟢 No risk - this is a self-contained piece of work

### DIFF
--- a/functions/actions/applications/applications.js
+++ b/functions/actions/applications/applications.js
@@ -82,6 +82,7 @@ module.exports = (config, firebase, db, auth) => {
     saveData[`_applications.${data.status}`] = firebase.firestore.FieldValue.increment(1);
     saveData['_applications._total'] = firebase.firestore.FieldValue.increment(1);
     saveData['_applications._lastUpdated'] = firebase.firestore.FieldValue.serverTimestamp();
+    if (data.personalDetails.fullName) saveData['_sort.fullNameUC'] = data.personalDetails.fullName.toUpperCase();
     await db.doc(`exercises/${data.exerciseId}`).update(saveData);
     console.log('success');
   }
--- a/functions/actions/applications/applications.js
+++ b/functions/actions/applications/applications.js
@@ -76,13 +76,20 @@ module.exports = (config, firebase, db, auth) => {
     console.log('application created');
     // slack.post(`${data.exerciseRef}. New application started`);
     if (data.userId) { await updateCandidate(data.userId); }
+
+    // update application
+    if (data.personalDetails.fullName) {
+      await ref.update({
+        '_sort.fullNameUC': data.personalDetails.fullName.toUpperCase(),
+      });
+    }
+
     // update counts
     console.log(`Update application counts: _applications.${data.status}`);
     const saveData = {};
     saveData[`_applications.${data.status}`] = firebase.firestore.FieldValue.increment(1);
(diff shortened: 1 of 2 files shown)


functions/actions/applications/applications.js
```diff
@@ -82,6 +82,7 @@ module.exports = (config, firebase, db, auth) => {
     saveData[`_applications.${data.status}`] = firebase.firestore.FieldValue.increment(1);
     saveData['_applications._total'] = firebase.firestore.FieldValue.increment(1);
     saveData['_applications._lastUpdated'] = firebase.firestore.FieldValue.serverTimestamp();
+    if (data.personalDetails.fullName) saveData['_sort.fullNameUC'] = data.personalDetails.fullName.toUpperCase();
```
warrensearle, line 85: This is updating the exercise document which is not what you want here. You'll need to update the `ref` passed in

functions/actions/applications/onUpdate.js
```diff
@@ -94,6 +94,25 @@ module.exports = (config, firebase, db, auth) => {
...
+    if (
+      !dataBefore._sort ||
+      (
+        dataAfter.personalDetails && dataAfter.personalDetails.fullName &&
+        dataBefore._sort.fullNameUC !== dataAfter.personalDetails.fullName.toUpperCase()
+      )
+    ) {
+      // update _sort.fullNameUC if fullName has changed
+      const data = {};
+      data['_sort.fullNameUC'] = dataAfter.personalDetails.fullName.toUpperCase();
```
warrensearle, line 107: This code could run if `dataBefore._sort` is missing however in this scenario we don't know whether we have `personalDetails.fullName` therefore you should check here, or in the `if` expression on line 99, that we have `personalDetails` and `personalDetails.fullName`

warrensearle: Actually modifying the `if` expression would be best as this update is only needed if we have `fullName`

warrensearle (changes requested):
Almost! 😄 
See comments

functions/actions/applications/applications.js
```diff
@@ -76,6 +76,14 @@ module.exports = (config, firebase, db, auth) => {
     console.log('application created');
     // slack.post(`${data.exerciseRef}. New application started`);
     if (data.userId) { await updateCandidate(data.userId); }
+
+    // update application
+    if (data.personalDetails.fullName) {
```
warrensearle, line 81: Please change to `if (data.personalDetails && data.personalDetails.fullName)` just to be sure 😄

warrensearle (changes requested):
Almost just one teeny tiny change

functions/actions/applications/applications.js
HalcyonJAC: Thank you for spotting it.

warrensearle (approved):
Great 👍

PR #823, opened 2023-01-12
Generate statutory consultation report

HalcyonJAC:
## What's included?
Add a callable function `generateStatutoryConsultationReport` to populate the statutory consultation report from the applications invited to the selection day.

## Who should test?
✅ Developers

## How to test?
Run the node script: `npm run nodeScript generateStatutoryConsultationReport`

### DIFF
--- a/functions/actions/exercises/generateStatutoryConsultationReport.js
+++ b/functions/actions/exercises/generateStatutoryConsultationReport.js
@@ -140,8 +140,9 @@ function getQualificationData(qualifications) {
     data[`qualificationLocation${index}`] = lookup(qualification.location) || '';
     data[`qualificationDate${index}`] = formatDate(qualification.date) || '';
     data[`completedPupillage${index}`] = helpers.toYesNo(qualification.completedPupillage) || '';
-    data[`notCompletePupillageReason${index}`] =
-      qualification.notCompletePupillageReason !== NOT_COMPLETE_PUPILLAGE_REASONS.OTHER ? lookup(qualification.notCompletePupillageReason) : qualification.details;
+    data[`notCompletePupillageReason${index}`] = qualification.completedPupillage ? '' : (
+      qualification.notCompletePupillageReason !== NOT_COMPLETE_PUPILLAGE_REASONS.OTHER ? lookup(qualification.notCompletePupillageReason) : qualification.details
+    );
   }
   return data;
 }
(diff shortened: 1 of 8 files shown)


functions/actions/exercises/generateStatutoryConsultationReport.js
```diff
@@ -0,0 +1,181 @@
...
+  return {
+    generateStatutoryConsultationReport,
+  };
+
+  async function generateStatutoryConsultationReport(exerciseId) {
+    // get submitted application with invited to selection day
+    const applicationRecords = await getDocuments(db.collection('applicationRecords')
+      .where('exercise.id', '==', exerciseId)
+      .where('stage', 'in', ['review', 'shortlisted', 'selected', 'recommended', 'handover'])
+    );
```
warrensearle, line 16: Comment suggests you are retrieving only applications with the 'invited to selection day' status however your code is retrieving all applications. Please check and update comment or code accordingly.

functions/shared/converters/lookup.js
```diff
@@ -206,6 +208,10 @@ const lookup = (value) => {
       // 'xxx': 'xxx',`
     };
 
+    lookup[NOT_COMPLETE_PUPILLAGE_REASONS.TRANSFERRED] = 'Qualified solicitor, qualified lawyer from another jurisdiction, or a legal academic transferred to the Bar';
+    lookup[NOT_COMPLETE_PUPILLAGE_REASONS.CALLED_PRE_2002] = 'Called to the Bar prior to 1 January 2002';
+    lookup[NOT_COMPLETE_PUPILLAGE_REASONS.OTHER] = 'Other - Please detail why you were exempt from undertaking pupillage by the Bar Standards Board';
```
warrensearle, line 213: Need to be careful here.
`NOT_COMPLETE_PUPILLAGE_REASONS.OTHER` has a value of "other" so this lookup will match for all occasions of 'other'.
I suggest removing the "Please detail why you..." text from the lookup. That feels like instruction text that should be on a form rather than part of a lookup.
Also check if there is already a constant for `'other'` and use that instead of `NOT_COMPLETE_PUPILLAGE_REASONS.OTHER`

warrensearle (changes requested):
Please see comments

functions/shared/converters/lookup.js
HalcyonJAC: Thank you for spotting that. The key `other` already exists and I will remove `NOT_COMPLETE_PUPILLAGE_REASONS.OTHER` in `lookup.js`.

functions/actions/exercises/generateStatutoryConsultationReport.js
HalcyonJAC: Thank you. I have updated the query to retrieve the applications whose status is `invitedToSelectionDay`.

warrensearle (approved):
👍

PR #850, opened 2023-03-07
admin#1912 Results should use z-score

warrensearle:
Updates results scoring to use average z score instead of average percentage, for qualifying tests

### DIFF
--- a/functions/shared/helpers.js
+++ b/functions/shared/helpers.js
@@ -22,7 +22,20 @@ module.exports = {
   removeHtml,
   normaliseNINs,
   normaliseNIN,
+  calculateMean,
+  calculateStandardDeviation,
 };
+
+function calculateMean(numArray) {
+  const mean = numArray.reduce((s, n) => s + n) / numArray.length;
+  return mean;
+}
+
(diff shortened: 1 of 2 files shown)


functions/shared/helpers.js
```diff
@@ -22,8 +22,21 @@ module.exports = {
...
 };
 
+function calculateMean(numArray) {
+  const mean = numArray.reduce((s, n) => s + n) / numArray.length;
+  return mean;
+}
+
+function calculateStandardDeviation(numArray) {
+  const mean = calculateMean(numArray);
+  const variance = numArray.reduce((s, n) => s + (n - mean) ** 2, 0) / (numArray.length - 1);
```
HalcyonJAC, line 36: Should here divide by `numArray.length` rather than `numArray.length - 1` here?

warrensearle: That's a good question. My understanding is there are a few versions of the standard deviation calculation: including `biased` and `unbiased`. When testing this I will check the calculations with our stats guy to confirm we are using the correct version

tomlovesgithub (approved):

PR #858, opened 2023-04-04
Admin#1948 Contacts export - additional field

HalcyonJAC:
Add additional fields to the contacts report:
- Welsh application


warrensearle (approved):
👍

PR #878, opened 2023-05-26
Restore document function

tomlovesgithub:
nodescript to restore documents when they have been moved to a deleted collection.


warrensearle (changes requested):
Looks useful. Can you move it into the `nodeScripts` folder rather than `nodeScripts/shared`. Shared is where config and other helper methods live, not the callable node scripts. 🤓

warrensearle (approved):
👍

PR #888, opened 2023-06-14
Release workflow: Move storage rules to last task

warrensearle:
There is currently a firebase issue preventing storage rules from deploying:
https://github.com/firebase/firebase-tools/issues/5955

Here I have moved our command to deploy storage rules to last in the list of tasks so that if it fails it doesn't prevent the other tasks from being run


tomlovesgithub (approved):

HalcyonJAC (approved):

joa178:
Hello!
I'm new at Firebase.
I have same issue, but I don't understand your solution. 
Can you explain me?
Thank you!

PR #920, opened 2023-08-08
Feature/875 search improvements

### LINKED ISSUE #875: Search improvements
We need to make some improvements to search across the platform:

- [x] Empty state pattern (see below)
- [x] Ensure the search works better (see the tuple pattern used in QT search)
- [x] Populate _search for existing collections using a nodeScript
- [x] Remove old search (eg computed->search for candidates) using a nodeScript
- [x] When performing a search if a valid search term has been passed in then override pagination to use the 'number' format. If the search term is removed or invalid then revert it back to whatever the initial value was.

---
**Empty state pattern**

An empty state pattern for when using a set of filters or searches return no results.

Currently we say "No Data" which is confusing to users

Image

Can we implement the example below to help users across all searches

https://www.gov.uk/search/all?keywords=dsadsa&order=relevance

Image

"When you leave the empty state empty, you are basically telling your users what there isn’t. You are missing out on the opportunity to tell them what there is.
For example, what could have been here, what they could gain from it, and what they can do to get things moving.”

Kinneret Yifrah
"People largely abandon applications for reasons which are within our control: because they do not perceive value from interacting with the software.” Patrick Mckenzie

**Improving Search**

Below is a list of collections and the fields we need to search on
- exercises - title and ref no
- independent assessments - assessor name + candidate name + email
- applications (currently a candidate search is used here then applications are searched by the candidates) -> simplifying by removing the candidate search ... candidate name + email + NI number + ref_no for application
- applicationRecords (same as above) ... this appears on the 'Stages' tab
- candidates - candidate name + email + NI number
- character checks - same as application search

drieJAC:
Closes #875


functions/shared/factories.js
```diff
@@ -323,7 +324,15 @@ module.exports = (CONSTANTS) => {
...
 
   function newAssessment(exercise, application, whichAssessor) {
+    // add search map
+    const search = getSearchMap([
+      application.personalDetails.fullName, // candidate name
+      application.referenceNumber,
+      // assessor details not available at this point!
+    ]);
+
     let assessment = {
```
warrensearle, line 334: Looks like this could be moved to the end of the `newAssessment` function at which point `assessment.assessor` field will have been populated

nodeScripts/temp/syncAssessmentsSearchMap.js
```diff
@@ -0,0 +1,71 @@
...
+    if (hasAssessorCandidateFullName) {
+      searchable.push(assessment.candidate.fullName);
+    }
+    if (hasApplicationReferenceNumber) {
+      searchable.push(assessment.application.referenceNumber);
+    }
+
+    commands.push({
+      command: 'update',
+      ref: db.collection('assessments').doc(assessment.id),
```
warrensearle, line 45: could use `assessment.ref` here

nodeScripts/temp/syncApplicationsSearchMap.js
```diff
@@ -0,0 +1,87 @@
...
+    }
+    if (hasReferenceNumber) {
+      searchables.push(application.referenceNumber);
+    }
+    
+    if (searchables.length) {
+      // Update application
+      commands.push({
+        command: 'update',
+        ref: db.collection('applications').doc(application.id),
```
warrensearle, line 46: could use `application.ref`

nodeScripts/temp/syncCandidatesSearchMap.js
```diff
@@ -0,0 +1,63 @@
...
+      searchable.push(candidate.computed.nino);
+    }
+
+    commands.push({
+      command: 'update',
+      ref: db.collection('candidates').doc(candidate.id),
+      data: {
+        _search: getSearchMap(searchable),
+        lastUpdated: firebase.firestore.FieldValue.serverTimestamp(),
+        'computed.search': firebase.firestore.FieldValue.delete(), // Remove nested field
```
warrensearle, line 40: 👍

warrensearle (changes requested):
Looks good to me, just a few tweaks. 👍

warrensearle (approved):
Looks great 👍 💯

PR #922, opened 2023-08-10
Hotfix To Fix Illegal Chars In Search Map

drieJAC:
Firestore does not support certain characters in it's keys and this is causing an error when users are searching in email addresses using the dot or @ symbol for example.
This fix replaces the illegal characters with a dash.


warrensearle (approved):
Looks good to me... and you have a unit test 👍

warrensearle (commented):
@drieJAC I notice similar functions have been added to JAC Kit. I wonder if we could use the JAC Kit functions here... what do you think?

PR #924, opened 2023-08-11
Hotfix To Fix Vacancy Vs Exercise Differences

drieJAC:
Node script to synchronise several fields that have become out of sync between exercise and vacancy which breaks the comparison between the two when showing changes during exercise approval process.
The vacancy fields need to adopt the value in the exercise.
The fields are:
aboutTheRoleWelsh, aSCApply, inviteOnly, previousJudicialExperienceApply, roleSummaryWelsh, welshPosts, welshRequirement


warrensearle (approved):
Looks good 👍

tomlovesgithub (approved):

PR #936, opened 2023-09-13
#1861 Amends to tasks, z score and status updates

warrensearle:
A few small tweaks to tasks, z score and status updates.

Part of jac-uk/admin#1861

### DIFF
--- a/functions/shared/config.js
+++ b/functions/shared/config.js
@@ -263,7 +263,7 @@ module.exports = {
   },
   TASK_TYPE: {
     SIFT: 'sift',
-    SELECTION: 'selection',
+    SELECTION_DAY: 'selectionDay',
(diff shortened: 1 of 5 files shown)


functions/shared/config.js
```diff
@@ -263,7 +263,7 @@ module.exports = {
   },
   TASK_TYPE: {
     SIFT: 'sift',
-    SELECTION: 'selection',
+    SELECTION_DAY: 'selectionDay',
```
HalcyonJAC, line 266: Should we also update `TASK_TYPE.SELECTION` in `nodeScripts/shared/config.js`?

warrensearle: Yes we should, thank you @HalcyonJAC

HalcyonJAC (approved):

PR #944, opened 2023-09-27
Outreach report fixes

drieJAC:
Reduce the footprint of the outreach report by selecting fields in the query, no longer fetching the application records and doing a single loop to build the application arrays.

Add the original keys back into the PAJE stats.
Add a nodeScript to add the _processing field to the application for those that are missing it. This pulls the stage and status into the application record so we dont need to query the application records when building the report anymore.

Closes jac-uk/admin#2097


HalcyonJAC (approved):

warrensearle (approved):
Nice. Thanks for optimising this 👍 

We'll need to run the migration script against develop and staging when this is merged... and again on prod before the next release

PR #952, opened 2023-10-10
[admin#1933] Remove QT code from digital platform - part II

warrensearle:
Removes remaining QT code from the digital platform (part II 😄)

Closes jac-uk/admin#1933


HalcyonJAC (changes requested):
Great work. I spotted there are some codes that might need to be removed as well:

- Line 3 in `functions/index.js` 
- Line 40 in `functions/index.js`
- `test/functions/updateStatus.spec.js`

PR #960, opened 2023-10-31
Bugfix/Admin 2158 live exercise export bug

HalcyonJAC:


nickaddy:
Sean has tested in the live environment: Checked this now Nick and can confirm it is working for me on the live platform.

drieJAC (commented):
All good

PR #987, opened 2023-12-19
#2252 User Management Comms & Search functionality

HalcyonJAC:
- Add a background function `onUserInvitationCreate` to send a notification email to the invited user.
- Add a background function `onUserCreate` to generate a user search map.
- Update the background function `onUserUpdate` to update the search map field in the user document.
- Update the background function `onUserDelete` to delete the user from auth DB only if the user has only one provider.
- Fix the bug when setting the user's custom claims.
- Update Firestore indexes to support the filter.
- Update the node script `populateJACUsers` to check the JAC email address from Microsoft and Google.


functions/actions/users.js
```diff
@@ -130,14 +133,22 @@ module.exports = (auth, db) => {
    */
   async function onUserUpdate(userId, dataBefore, dataAfter) {
     const fields = ['displayName', 'email', 'disabled'];
-    const data = {};
+const data = {};
```
warrensearle, line 136: Code needs indenting 🤓

warrensearle (changes requested):
@HalcyonJAC I don't think we should be parsing the display name, trying to infer first name and last name and then storing the results in our database.

Let's just use display name as it comes through from Microsoft. This appears to be "lastName, firstName" so if we need sorting by lastName then we already have that. Likewise search can be based on email and display name, so we don't need to have parsed the display name for that either.

HalcyonJAC:
@warrensearle Thanks for your feedback. I guess most admin users will log in to the platform with their Microsoft account and fewer with their Google account. The format of the display name from Google will be like "firstName lastName" which is different from Microsoft. I am not sure if we need to consider the display of users from Google.

nickaddy:
@HalcyonJAC In terms of testing, if I create a new account on production, the user should receive an email confirmation, right?

HalcyonJAC:
> @HalcyonJAC In terms of testing, if I create a new account on production, the user should receive an email confirmation, right?

@nickaddy Yes, the user should receive an invitation confirmation email if you create a new user invitation.

warrensearle (approved):
Thanks @dodosoya looks good to me

PR #1048, opened 2024-03-15
Include permission to configure exercise

warrensearle:
Adds a new permission to enable users to configure an exercise - as in change processing version number and/or application form version number.

Note: Firestore security rules have not been updated to check for this permission. Currently the permission is solely used in the UI


HalcyonJAC (approved):
👍

PR #1049, opened 2024-03-15
Bugfix/admin 2275 Generate Diversity Report

HalcyonJAC:
- Fix the count when generating the diversity report.


warrensearle (approved):
👍

PR #1064, opened 2024-03-26
Feature/admin 2356 Improvements to the Diversity Report

HalcyonJAC:
- Add the additional data to the diversity report.

### DIFF
--- a/functions/actions/exercises/generateDiversityReport.js
+++ b/functions/actions/exercises/generateDiversityReport.js
@@ -1,6 +1,7 @@
 const { getDocument, getDocuments } = require('../../shared/helpers');
 const { applicationOpenDatePost01042023 } = require('../../shared/converters/helpers');
 const { availableStages } = require('../../shared/exerciseHelper');
+const config = require('../../shared/config');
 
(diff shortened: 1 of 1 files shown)


functions/actions/exercises/generateDiversityReport.js
```diff
@@ -1,6 +1,7 @@
 const { getDocument, getDocuments } = require('../../shared/helpers');
 const { applicationOpenDatePost01042023 } = require('../../shared/converters/helpers');
 const { availableStages } = require('../../shared/exerciseHelper');
+const config = require('../../shared/config');
```
warrensearle, line 4: Hey @HalcyonJAC currently our constants and config variables are together in one file, called `config`. Given this, please could you inject this in rather than importing directly here as it could get messy/confusing if in this module we also needed to use the config variables. The code would not behave as expected when run via nodeScript. Thank you!

(the params on line 24/25 is where to receive the injected `config` object)

warrensearle (changes requested):

drieJAC (approved):

PR #1073, opened 2024-04-18
Nodescript to update return and hard limit dates for IAs

drieJAC:
Added nodescript to update return and hard limit dates for IAs.

This is based on a request in this ticket raised by Foyez Miah: https://github.com/jac-uk/ticketing-system/issues/287

### DIFF
--- a/nodeScripts/updateAssessmentDates.js
+++ b/nodeScripts/updateAssessmentDates.js
@@ -1,0 +1,90 @@
+/*
+ * This script is used to update the assessment due date/hard limit dates
+ * It updates the dates in the exericse and the assessments
+ * Optionally pass either or both of the following (set to null otherwise)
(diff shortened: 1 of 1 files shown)


nodeScripts/updateAssessmentDates.js
```diff
@@ -0,0 +1,90 @@
+/*
+ * This script is used to update the assessment due date/hard limit dates
+ * It updates the dates in the exericse and the assessments
+ * Optionally pass either or both of the following (set to null otherwise)
+ */
+
+'use strict';
+
+const config = require('./shared/config');
+const { firebase, app, db } = require('./shared/admin');
```
HalcyonJAC, line 10: Could we remove `config` and `firebase` as they are not used in this file?

PR #1087, opened 2024-05-13
Feature/TS 307 Remove Historic Status in ApplicationRecord 

HalcyonJAC:
Closes [User Raised Issue BR_ADMIN_PR_000110 #307](https://github.com/jac-uk/ticketing-system/issues/307)

- Add a node script `removeApplicationRecordStatus` to remove the specified status in the applicationRecords.

### DIFF
--- a/nodeScripts/removeApplicationRecordStatus.js
+++ b/nodeScripts/removeApplicationRecordStatus.js
@@ -19,7 +19,7 @@ const main = async () => {
 
   // get applicationRecords with specified status for the exercise
   console.log('-- Fetching applicationRecords...');
-  const ref = db.collection('applicationRecords').where('exercise.id', '==', exerciseId);
+  const ref = db.collection('applicationRecords').where('exercise.id', '==', exerciseId).select('status', 'statusLog');
   const applicationRecords = await getDocuments(ref);
   console.log(`-- Fetched applicationRecords: ${applicationRecords.length}`);
   const commands = [];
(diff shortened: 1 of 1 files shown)


KoWeiJAC (approved):

nodeScripts/removeApplicationRecordStatus.js
```diff
@@ -0,0 +1,68 @@
...
+const status = '';
+
+const main = async () => {
+  if (!exerciseId || !status) {
+    throw new Error('Please specify exerciseId and status');
+  }
+
+  // get applicationRecords with specified status for the exercise
+  console.log('-- Fetching applicationRecords...');
+  const ref = db.collection('applicationRecords').where('exercise.id', '==', exerciseId);
```
warrensearle, line 22: @HalcyonJAC is there a way to update the firestore query to only return the documents we are interested in rather than having to get them all? (I am guessing not as this would require a new index... but thought I'd ask anyway!)

warrensearle (approved):
Looks good, just a question about whether the query could be optimised - see comments.
👍

nodeScripts/removeApplicationRecordStatus.js
HalcyonJAC: The query could be optimised like:

```
const ref = db.collection('applicationRecords').where('exercise.id', '==', exerciseId).orderBy(`statusLog.${status}`).select('status', 'statusLog');
```
If we do it this way, we might need to add a new index based on the specified status `statusLog.otherPassed` which would not be a good solution. I am not sure if there is another way to optimise it.

PR #1098, opened 2024-05-29
Added checks for prop changes that account for undefined attrs.

tomlovesgithub:
Only adding values to the search map if they exist.


tomlovesgithub:
- [ ] 

- [ ]

PR #1105, opened 2024-06-10
2408/self assessment panel packs

tomlovesgithub:
Changes to getHtmlPanelPack that will add the extracted content to the file if there is content directly on the application (new way) or in cases where self assessment has been done the old way, the self assessment itself will be copied to the folder as before.

closes #2408

### DIFF
--- a/functions/shared/converters/applicationConverter.js
+++ b/functions/shared/converters/applicationConverter.js
@@ -49,7 +49,7 @@ module.exports = () => {
         html.addTable(additionalPrefs);
       }
       
-      if (application.uploadedSelfAssessment && exercise.selfAssessmentWordLimits) {
+      if (application.uploadedSelfAssessment) {
         const selfAssessment = getSelfAssessment(application, exercise);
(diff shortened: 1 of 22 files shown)


HalcyonJAC (approved):

functions/shared/converters/applicationConverter.js
```diff
@@ -48,6 +48,14 @@ module.exports = () => {
         html.addHeading('Additional Preferences');
         html.addTable(additionalPrefs);
       }
+      
+      if (application.uploadedSelfAssessment && exercise.selfAssessmentWordLimits) {
```
drieJAC, line 52: I think it would be better to put the check for exercise.selfAssessmentWordLimits into getSelfAssessment.
Will make the getSelfAssessment function a bit more reusable.

tomlovesgithub: made some changes, let me know if thats what you were thinking of!

drieJAC (approved):

PR #1143, opened 2024-07-17
Feature/admin 2492 Transfer NINos from candidates to their applications on JAC00203

HalcyonJAC:
- Add a node script to copy the national insurance number from the candidate document to the application document.


tomlovesgithub (approved):
👍

PR #1160, opened 2024-08-01
Bugfix/Fix NPM Vulnerabilities

HalcyonJAC:
## What's included?

[Identify And Fix NPM Vulnerabilities In All Repositories #2513](https://github.com/jac-uk/admin/issues/2513)

## Who should test?
✅ Developers

## How to test?

Run `npm audit` in the terminal and check if there are no high vulnerabilities.

### DIFF
--- a/functions/package-lock.json
+++ b/functions/package-lock.json
@@ -8589,9 +8589,9 @@
       }
     },
     "node_modules/notifications-node-client/node_modules/axios": {
-      "version": "1.7.2",
-      "resolved": "https://registry.npmjs.org/axios/-/axios-1.7.2.tgz",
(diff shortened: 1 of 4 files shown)


warrensearle (commented):
Are you sure this is correct? Seems strange to go back to allowing version 0.0.1 and above when previously it was version 4.2.6 and above!

tomlovesgithub (changes requested):
Can we discuss this change at the next dev meeting? seems a little sketchy to downgrade to such an early version

tomlovesgithub (approved):

warrensearle (approved):

PR #1174, opened 2024-08-19
Merit list changes

warrensearle:
Some changes to tasks which use panels

Resolves jac-uk/admin#2247


functions/actions/tasks/meritListHelper.js
```diff
@@ -0,0 +1,160 @@
...
+  const match = getOverrideReasons().find(reason => reason.code === code);
+  return match ? match.value : '';
+}
+
+function convertOverrideValue2Code(value) {
+  const match = getOverrideReasons().find(reason => reason.value === value);
+  return match ? match.code : '';
+}
+
+function isPassingScore(task, score) {
```
HalcyonJAC, line 68: Should we return false if no condition is met?

warrensearle: Yes we should 👍

PR #1179, opened 2024-08-28
Added a nodeScript which prompts the user for feedback

drieJAC:
This script is an example of how we can prompt users for feedback when running nodescripts.
We can confirm they want to run the script vs the environment they are on (for instance production).


KoWeiJAC (approved):
Good idea!

HalcyonJAC (approved):

tomlovesgithub (approved):
🫡

PR #1206, opened 2024-09-25
Feat/admin 2571 generate a list of candidates for comms

KoWeiJAC:
closes [Admin-2571](https://github.com/jac-uk/admin/issues/2571)

### DIFF
--- a/nodeScripts/listCandidateEmailsForComm.js
+++ b/nodeScripts/listCandidateEmailsForComm.js
@@ -1,0 +1,122 @@
+'use strict';
+
+import { firebase, app, db } from './shared/admin.js';
+import { getDocuments } from '../functions/shared/helpers.js'
+import _ from 'lodash'
+import { createRequire } from 'module';
+const require = createRequire(import.meta.url);
+
+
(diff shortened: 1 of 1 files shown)


nodeScripts/listCandidateEmailsForComm.js
```diff
@@ -0,0 +1,134 @@
...
+ *   npm run nodeScript listCandidateEmailsForComm
+ *   ```
+ */
+
+'use strict';
+
+import { db } from './shared/admin.js';
+import _ from 'lodash'
+import { createRequire } from 'module';
+const require = createRequire(import.meta.url);
```
HalcyonJAC, line 18: Do we need this line?

KoWeiJAC: @HalcyonJAC 
Good spot! This is for using require syntax, but we can use the import syntax instead!
I just pushed a change for it.

HalcyonJAC (approved):

tomlovesgithub (approved):

PR #1243, opened 2024-11-01
dependencies oct24

tomlovesgithub:


warrensearle:
> @warrensearle are you happy with using "overrides"?

Be good to look at what dependencies use the dependency you are overriding in case we can use a newer patched version. If not and this override avoids a vulnerability and is non-breaking then go for it 👍 
(we should make sure we review overrides)

warrensearle (approved):
This is some awesome tidying @tomlovesgithub. Zero vulnerabilities! 💪 

The code changes look good but as this is a major upgrade we will need to ensure the platform gets thoroughly tested before we roll out to production.

Let's get this merged and deployed to develop & staging and then let testing commence! 💯 

@nickaddy FYI

HalcyonJAC (approved):

PR #1263, opened 2024-11-18
1261 Raise An Issue Not Always Sending Slack Alerts

### LINKED ISSUE #1261: Raise An Issue Not Always Sending Slack Alerts
## Background
When certain actions are performed on Github (in the Ticketing System Repo, a webhook is triggered which calls an endpoint in the Digital Platform passing data to it. This call subsequently creates a Slack alert in a dedicated channel to alert the developers of an issue raised on the platform.
From time to time this call is timing out or failing for a related reason.

## User Story
As a developer I would like a scheduled task to check whether a slack notification has been sent to inform us of a new bug and if not for the Digital Platform to trigger a Slack alert.
The data in the bugReport record will have this structure:

slackMessages = {
      onCreate: {
        retries: 0,
        sentAt: null,
      },
    };

## Benefit(s) to user (if not already clear from User story)
We always receive Slack alerts when a new issue is created/assigned

## Feature(s) Description

- [x] ADMIN: When a bugReport record is created in firebase add a timestamp to the record to track when the slack message was sent on creation of the issue. This timestamp should be set to null. Also add a field called retries and set it to zero.
- [x] Ensure the timestamp gets updated when the github webhook calls the Digitial Platform
- [x] Create a scheduled task in the Digital Platform which runs every 2 mins and checks for bugReports with the timestamp field set to null and the retries value less than 4. For these records we should increment the retries first, then send a slack message, then on successful completion set the timestamp.

## Acceptance Criteria 
It's done when the slack message is either sent every time a bug issue is raised or there are 3 failing attempts.

## Definition of Done
- [ ] User stories/acceptance criteria met
- [ ] Internal reviews passed (feedback actioned)
- [ ] User testing passed
- [ ] Relevant technical documentation updated
- [ ] User guidance updated
- [ ] Deployed and merged without errors

drieJAC:
Occasionally the slack message doesnt get sent when the create issue github hook calls our backend.
This PR allows us to track whether this message was sent and, if not, retry sending the message several times (specified by a constant in the config).
Added a scheduled task which runs every 2 mins to look for bugReports which havent had a slack message sent. If it finds any it tries to send the slack message.
Also modified existing code to:
- ensure the github link to the issue and the issue number are stored in the bugReport
- no longer send the slack message when the 'create issue' hook is called and leave it to the scheduled task to pick it up and send it

Closes #1261

### DIFF
--- a/functions/actions/zenhub/sendNewIssueSlackNotifications.js
+++ b/functions/actions/zenhub/sendNewIssueSlackNotifications.js
@@ -1,0 +1,123 @@
+import initSlack from '../../shared/slack.js';
+import initUsers from '../users.js';
+import initBugReports from '../bugReports.js';
+import get from 'lodash/get.js';
+
+export default (config, db, auth, firebase) => {
+
+  const slack = initSlack(config);
+  const { getUser } = initUsers(auth, db);
+  const { incrementSlackRetries, updateBugReportOnSlackSent } = initBugReports(db, firebase);
+
+  return {
+    sendNewIssueSlackNotifications,
+  };
+
+  /**
+   * Called for new issues (or those previously created whose slack messages weren't sent out)
+   * Send a message to a Slack channel notifying the team of the bug
+   * Increment retries then send a slack message and only if successful set the timestamp
+   * 
+   * @param {object} params 
+   * @param {object} bugReport
+   * @param {string} slackChannelId
+   */
+  async function sendNewIssueSlackNotifications(bugReport, slackChannelId) {
+    const reporterUser = await getUser(bugReport.userId);
+
+    // Build the slack message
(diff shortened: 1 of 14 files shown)


functions/actions/zenhub/sendNewIssueSlackNotifications.js
```diff
@@ -0,0 +1,123 @@
...
+    else {
+      console.log(`Slack request failed for bugReport: ${bugReport.referenceNumber}`);
+      return false;
+    }
+  }
+
+  function addSlackDivider() {
+    return {
+			'type': 'divider',
+		};
```
HalcyonJAC, line 57: Could you fix the indent, please?

PR #1288, opened 2024-12-17
Merit MVP further changes

warrensearle:
Further amends following the pilot testing of the panellist app, grade sheet and merit list.

 - Ensure there are applications for a Task before creating it
 - Remove withdrawn candidates from a task, rather than marking them as optional
 - Allow panellists to read & update their own data


nickaddy:
@warrensearle I think we've found an alternative solution for this, right?

warrensearle:
Thanks @nickaddy we certainly have. I'll make those changes on this PR.

tomlovesgithub (approved):

PR #1297, opened 2025-01-06
1279 TECH DEBT - Reduce footprint of callable functions as reaching limit

drieJAC:
### THIS IS CURRENTLY A WORK IN PROGRESS SO DON'T CODE REVIEW YET

Rejigged callable functions so as many as possible are now imported from modules


drieJAC:
Moving the

drieJAC:
This was testing an approach that didnt work

PR #1299, opened 2025-01-09
2659 generate list of active users

drieJAC:
Generate a list of active users who signed in on or after 01/01/2024 and export to csv.
Closes jac-uk/admin#2659


HalcyonJAC (changes requested):
Could we remove the file `nodeScripts/temp/candidates-signin-since-01-01-2024.xlsx` from the codebase as it might contain sensitive data, please?

functions/actions/internalReporting.js
```diff
@@ -81,5 +84,121 @@ export default (db, auth) => {
...
+      const users = [];
+      await listAllUsers(users);
+      for (const user of users) {
+        let isJacAdmin = false;
+        const cutoffDate = new Date(specifiedCutoffDate); // Set cutoff date
+        const lastSignInTime = user.metadata.lastSignInTime;
+        if (lastSignInTime && new Date(lastSignInTime) >= cutoffDate) {
+          if (user.providerData.length === 1) {
+            const provider = user.providerData[0];
+            if (user.email.match(/(.*@judicialappointments|.*@justice)[.](digital|gov[.]uk)/) &&
```
KoWeiJAC, line 108: May can extract a isJacAdmin helper function that can be used.

functions/actions/internalReporting.js
```diff
@@ -81,5 +84,121 @@ export default (db, auth) => {
...
```
KoWeiJAC, line 165: May can extract a isJacAdmin helper function that can be used.

KoWeiJAC (approved):

HalcyonJAC (approved):

nickaddy:
@drieJAC Thanks Omar, looking good. Column AE 'Same as sex registered at birth' seems to be working fine, but can you address the following pls?

- [ ] Rename column Q in download to 'Country' - the content should come from the Citizenship field as most candidates will not specify country in their address
- [ ] Data is missing in Columns Q or V - I've checked the applications and data is present
- [ ] Data is missing in Columns AM and AN - I've checked the applications and data is present
- [ ] Column D 'Candidate Last Name' appears to be referencing the 'Candidate title' field
- [ ] Rename Column BC 'Suitable for vacancies in Wales?' and remove data link - this column will be populated manually by SETs
- [ ] Names not appearing onscreen on Admin (you're already aware of this)

PR #1300, opened 2025-01-10
Fix firebase functions upgrade

tomlovesgithub:
specify version 1 import for the firebase-function as per v5.** > v6.** upgrade


warrensearle (approved):
👍

PR #1330, opened 2025-02-15
1279 TECH DEBT - Reduce footprint of callable functions as reaching limit

### LINKED ISSUE #1279: TECH DEBT - Reduce footprint of callable functions as reaching limit
## Background
We are reaching the limit of the number of callable functions we can deploy. This makes deploying them slower and error prone.

## User Story
We need to reduce the footprint of callable functions. Deploy them as modules.

drieJAC:
### Description
Upgraded eslint ecma version to support gen 2 functions.
Modified **callable, scheduled and http functions to gen 2**.
Moved secrets into secrets mgr so now pulling them from secrets mgr in the top level function and passing them down into actions.
Fixed eslint warnings.

Closes #1279 

### Config
The config values now belong in either the `/functions/.env.{ENVIRONMENT}` file (for non-sensitive values) or the secrets manager (for sensitive values). There are instructions below for how to add secrets to the secret mgr.
If you modify any of these config values you will need to redeploy the functions that use them or just redeploy all functions. However **before deploying any of the gen 2 functions the existing gen 1 function needs to be deleted**, as per the instructions below.

### Live Deployment Instructions
When deploying to production the following steps need to be taken:
- Go to the firebase secrets manager for the production environment and enable the secrets manager
- Add the following keys and their values:

1. GITHUB_PAT
2. GITHUB_WEBHOOK_SECRET
3. GOOGLE_RECAPTCHA_SECRET
4. NOTIFY_KEY
5. QT_URL
6. QT_KEY
7. SLACK_URL
8. SLACK_CHANNEL_ID
9. SLACK_TICKETING_APP_BOT_TOKEN
10. ZENHUB_GRAPH_QL_API_KEY
11. ZENHUB_ISSUES_WORKSPACE_ID

Most of these values can be found in the firebase functions config for the project in question.
To create a secret use the following command:

> firebase functions:secrets:set < key > --project=< project ID >

eg 

> firebase functions:secrets:set NOTIFY_KEY --project=platform-production-9207d

You will then be prompted for the value.
This needs to be done for each key above

- You can then remove the config keys from firebase functions config by switching to that project:

> firebase use < project ID >

Then remove the key using the following command:

> firebase functions:config:unset < key >

eg

> firebase functions:config:unset zenhub.workspace_id

- **Also these keys should NOT exist in the .env file as this will cause errors!**
- **Delete the functions** (this can be done from the command line or in the firebase console where the functions are (click the ellipsis to the right of each function)
- Deploy the functions:

> firebase deploy --only functions

### Callable Functions In Apply

The following is a list of callable functions in the Apply project. They need extra attention when testing to ensure they work properly. I have tested them on DEVELOP and they are working fine:

- sendSmsVerificationCode
- verifySmsVerificationCode
- logEvent
- verifyRecaptcha
- checkEnabledUserByEmail
- extractDocumentContent
- updateEmailAddress


warrensearle:
Hey @tomlovesgithub @KoWeiJAC @HalcyonJAC I have made a number of tweaks to the original changes from @drieJAC.

- Removed use of `defineSecret` and associated boilerplate as we can simply use strings instead
- Removed secrets & env values from module parameters in favour of using `process.env.{KEY}` directly where it is needed
- Removed env values from `config.js` and renamed to `constants.js` instead
- Upgraded to use node 20
- Converted background functions to 2nd generation
- Added `.env.develop` and `.env.local` to source control and moved a few private-ish values to secrets instead

warrensearle (approved):
Approving as I'd like us to merge this in so we can test on develop and staging (just need to add the `.env.staging` file for this and set up the secrets)

As I have made a lot of changes on this PR it would be great to get reviews from @HalcyonJAC @KoWeiJAC and @tomlovesgithub 🙏

KoWeiJAC (approved):
👍👍👍

HalcyonJAC (approved):

